PR #12, opened 2020-01-07
Don't insert space if RBUFFER already starts with space

romkatv:
Example:

```text
gre█ 'query'
   ^ cursor
```
Press TAB and select `grep` from fzf. With `FZF_TAB_INSERT_SPACE=1` this used to result in the following completion:

```text
grep  'query'
    ^^ two spaces
```

With this PR:

```text
grep 'query'
    ^ one space
```

While at it, handle `FZF_TAB_INSERT_SPACE=0` as an explicit instruction to NOT insert space.


Aloxaf:
Thank you very much!

romkatv:
Thanks for the super fast merge :100:

PR #13, opened 2020-01-07
Allow selecting multiple items

### LINKED ISSUE #11: [FR] Allow selecting multiple items
Sorry for spamming the issue tracker, this is the last one for a while I promise

FZF allows selecting multiple items using the tab key. I don't know the ins and outs of this plugin but I suspect that the tab key is captured by ZSH and not passed to FZF.

There is a conflict because most people probably expect to be able to cycle completions using the tab key, so could we maybe use ctrl-tab or ctrl-return?

Aloxaf:
Fix #11, use ctrl-space to select.


Aloxaf:
Outdated

PR #14, opened 2020-01-07
Minor stylistic improvements in parameter declarations

romkatv:
1. Use `: ${...}` instead of `[[ ${...} ]]` as this is the standard zsh idiom for discarding values of expansions performed for their side effects.
2. It's valid for `FZF_TAB_OPTS` to have empty value, so it shouldn't be replaced with the default.
3. Initialize `FZF_TAB_QUERY` the same way as other parameters for the sake of consistency.

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -55,12 +55,10 @@ function compadd() {
     builtin compadd -Q -U ''
 }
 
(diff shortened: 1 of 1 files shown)


Aloxaf:
Learned a new syntax. Thanks a ton!

PR #15, opened 2020-01-07
Minor performance improvements

romkatv:
Speed things up a little:

- Avoid forks.
- Avoid loops.

The standard way to return values from zsh functions (be it scalars, arrays or associative arrays) is through named parameters. This works thanks to dynamic scope.

```zsh
# "returns" $1 repeated 3 times
function f() { typeset -g REPLY=$1$1$1 }

function g() {
  local REPLY
  f foo
  echo -E - $REPLY
}

REPLY=bar
g                 # prints "foofoofoo"
echo -E - $REPLY  # prints "bar"
```

There is one non-trivial combination of expansions in this PR that is used to find common prefix. I added several comments around it to make it less impenetrable. Hopefully they make the code understandable.


Aloxaf:
Thank you very much!

PR #16, opened 2020-01-08
Survive any combination of zsh options and global aliases

romkatv:
There are two parts to this commit:

1. Ensure that `fzf-tab.zsh` is always parsed the same way regardless of options and global aliases. For example, if `alias -g input='<$TTY'` is defined before `fzf-tab.zsh` is sourced, it results in an error without this commit. The solution is to unset `aliases` at the top of `fzf-tab.zsh` and set it back at the bottom. Until this option is unset, everything has to be quoted to prevent alias expansion (hence `'builtin' 'local' '-a' '_fzf_tab_opts'`). Similar measures are taken w.r.t. `sh_glob` and `brace_expand` as these options also affect how functions are parsed.
2. Ensure that functions execute with a known set of options. For example, the code relies in several places on array subscripts being 1-based. If the user sets `ksh_arrays`, the code will break. The solution is to call `emulate -L zsh` at the beginning of all functions that may be invoked with user-defined options. There is one exception: `_main_complete` must be called with user-defined options because its results are option-dependent. For example, `_main_complete` ignores files whose names begin with `.` unless `dot_glob` is set.

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -1,7 +1,15 @@
+# temporarily change options
+'builtin' 'local' '-a' '_fzf_tab_opts'
+[[ ! -o 'aliases'         ]] || _fzf_tab_opts+=('aliases')
+[[ ! -o 'sh_glob'         ]] || _fzf_tab_opts+=('sh_glob')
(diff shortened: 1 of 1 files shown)


Aloxaf:
Thanks a lot!

PR #17, opened 2020-01-08
Minor cleanup

romkatv:
1. Make `i` local.
2. Stop setting `fzf_tab_default_completion` as it's no longer used.
3. Restore the original `^I` widget in `disable-fzf-tab`.


Aloxaf:
:+1:

PR #22, opened 2020-01-09
Allow selecting multiple items

### LINKED ISSUE #11: [FR] Allow selecting multiple items

Aloxaf:
Fix #11 
Select with `Ctrl+Space`

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -58,8 +58,8 @@ compadd() {
         fi
         compcap[$dscr]=$__tmp_value${word:+$'\0'"word"$'\0'$word}$'\0'"args"$'\0'${(pj:\1:)_opts}
     done
-    # tell zsh that the match is successful
-    builtin compadd -Q -U ''
+    # hack: pretend to have done a successful compadd ( see _alternative:76 )
+    nm=-1
 }
 
 : ${FZF_TAB_INSERT_SPACE:='1'}
@@ -129,7 +129,7 @@ _fzf_tab_get_candidates() {
 
 _fzf_tab_complete() {
     local -A compcap
-    local choice
+    local choice choices
 
     IN_FZF_TAB=1
     _main_complete  # must run with user options; don't move `emulate -L zsh` above this line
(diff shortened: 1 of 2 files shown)


fzf-tab.zsh
```diff
@@ -59,12 +59,20 @@ compadd() {
...
+# FIXME: what if the left char is not whitespace: `echo $widgets[\t`
+_fzf_tab_remove_space() {
+    [[ LBUFFER[-1] == ' ' ]] && LBUFFER[-1]=''
 }
 
 : ${FZF_TAB_INSERT_SPACE:='1'}
 : ${FZF_TAB_COMMAND:='fzf'}
-: ${FZF_TAB_OPTS='--cycle --layout=reverse --tiebreak=begin --bind tab:down,ctrl-j:accept --height=15'}
+: ${FZF_TAB_BIND:='tab:down,ctrl-j:accept,ctrl-space:toggle'}
+: ${FZF_TAB_OPTS="--cycle --layout=reverse --tiebreak=begin -m --bind $FZF_TAB_BIND --height=15"}
```
romkatv: This disallows empty bindings. It would be nice to allow users to set `FZF_TAB_BIND=""` to explicitly signal that they don't want custom bindings.

Alternatively, don't introduce `FZF_TAB_BIND` at all. This kind of parameter creep can get messy pretty fast. If starting over, I would even get rid of `FZF_TAB_OPTS` and make `FZF_TAB_COMMAND` an array parameter. This is harder to do now as there are already users of this code.

ReekyMarko:
I just wanted to let you know that I tested out this branch, everything seems to work fine :smile:

fzf-tab.zsh
Aloxaf: But fzf don't allow empty bindings, so it must be a non-empty value.

I agree with you. In fact, I don't plan to expose `FZF_TAB_BIND` to users, I just want to make the code cleaner. 
Making it an array parameter is a nice idea. I didn't expect that the options will be so long...

romkatv: > But fzf don't allow empty bindings, so it must be a non-empty value.

What I meant is that some users (myself included) might prefer to have no custom bindings in fzf. The code appears to expose `FZF_TAB_BIND` as a parameter for defining custom bindings, but it doesn't allow saying "no custom bindings". A natural fix would look something like this:

```zsh
: ${FZF_TAB_BIND='tab:down,...'}
: ${FZF_TAB_OPTS="... ${FZF_TAB_BIND:+--bind $FZF_TAB_BIND}"}
```

A single array would be better of course. It would be even better if fzf-tab didn't modify parameters that are meant to be set by end users.

Aloxaf:
@ReekyMarko Thanks for your test. 
But it can't deal with the situation where the character to the left of the cursor is not whitespace. And I can't come up with a better solution. So this PR may not be merged in a short time.

ReekyMarko:
@Aloxaf oh I didn't notice that. I hope you will find a solution

PR #23, opened 2020-01-09
Support grouping

### LINKED ISSUE #19: Duplicate completion candidates
To reproduce:

1. Install zsh 5.7.1 (I haven't tried any other version).
2. Run `git clone https://github.com/Aloxaf/fzf-tab.git ~/fzf-tab`.
3. Run `zsh -df`.
4. Run `autoload compinit && compinit`.
5. Run `cd ~/fzf-tab`.
6. Type `git checkout maste` and hit `TAB`.
7. Observe `maste` being completed to `master`.
8. Hit Ctrl-C and run `source ~/fzf-tab/fzf-tab.zsh`.
9. Type `git checkout maste` and hit `TAB`.
10. Observe the **unexpected** appearance of fzf:
```text
adam% git checkout maste
> master                                                                              
  2/2                                                                                 
> master                                                                              
  master  -- Minor cleanup (#17)                                                      
```

Aloxaf:
This PR gives different colors to different groups, and keep the group order.

Fix #19 

TODO:
- [x] Minimal implementation
- [x] Put multiple header in one line
- [x] Hide the group which has no members
- [x] Better default color?
- [x] Make the code cleaner

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -7,8 +7,6 @@
 
 zmodload zsh/zutil
 
-autoload -U colors; colors
-
 # thanks Valodim/zsh-capture-completion
 compadd() {
     emulate -L zsh
@@ -34,6 +32,9 @@ compadd() {
     if (( $#__hits == 0 )); then
         return
     fi
+
+    # keep order of group description
+    _fzf_tab_groups+=$expl
 
     # store these values in _fzf_tab_compcap
     local -a keys=(apre hpre isfile expl PREFIX SUFFIX IPREFIX ISUFFIX)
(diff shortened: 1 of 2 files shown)


fzf-tab.zsh
```diff
@@ -158,8 +185,9 @@ _fzf_tab_complete() {
     if [[ -n $choice ]]; then
         local -A v=("${(@0)${_fzf_tab_compcap[$choice]}}")
         local -a args=("${(@ps:\1:)v[args]}")
+        [[ -z $args[1] ]] && args=()  # don't pass an empty string
         IPREFIX=$v[IPREFIX] PREFIX=$v[PREFIX] SUFFIX=$v[SUFFIX] ISUFFIX=$v[ISUFFIX] \
-               builtin compadd "${args[@]:-Q}" -Q -- $v[word]
+               builtin compadd "${args[@]:--Q}" -Q -- $v[word]
```
romkatv: Is double `-Q` intentional?

fzf-tab.zsh
```diff
@@ -123,15 +128,30 @@ _fzf_tab_get_candidates() {
...
+        else
+            candidates+=1$'\b\0'$k$'\0'$dsuf
+        fi
     done
     (( same_word )) && candidates[2,-1]=()
     local LC_ALL=C
     candidates=("${(@on)candidates}")
+
+    for k in {1..$#_fzf_tab_groups}; do
+        headers+=$FZF_TAB_GROUP_COLOR[k]$_fzf_tab_groups[k]$reset_color
```
romkatv: You need to autoload and call `colors` for `reset_color` to become available. However, it's better to hard-code $\033[00m` since you've hard-coded the other escape codes anyway.

Aloxaf: Oh, I removed `colors` in the second commit but forgot to expand it. Thanks.

fzf-tab.zsh
Aloxaf: Yes. I should also fix it in master. It's expected to expand to `-Q` when `args` is empty, or `compadd` will treat it as an empty candidate.

PR #26, opened 2020-01-16
minor fixes

romkatv:
1. Use `function foo() {..}` syntax for defining functions instead of `functions[foo]='...'`.
2. Use unique and ugly names for variables that must be reachable across scopes outside of fzf-tab's control.
3. Don't use `&&` for control flow when executing code with user options. It breaks when `err_return` is set.
4. Use `always` to restore `_main_complete` to make breakage less likely.


Aloxaf:
Thanks a lot!

PR #28, opened 2020-02-06
Continuous completion

Aloxaf:
Use '/' to trigger

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,10 +42,10 @@ git clone https://github.com/Aloxaf/fzf-tab ~ZSH_CUSTOM/plugins/fzf-tab
 
 ## Usage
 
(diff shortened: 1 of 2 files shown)


romkatv:
Would it be possible to use a different binding by default? Perhaps `alt-enter`? With the current binding it's no longer possible to type `/` in fzf.

nasyxx:
I guess there should be no "/" in the file name, which makes it is exact to use `/` as the default binding.

romkatv:
> I guess there should be no "/" in the file name, which makes it is exact to use `/` as the default binding.

Completion candidates can contain "/" both in their values and in descriptions.

Aloxaf:
I will make it configurable.

romkatv:
Thank you!

PR #35, opened 2020-02-17
Allow to keep the original order of candidates

### LINKED ISSUE #29: Show identical keys nearby
Output of paired parameters nearby as in the **fzf-tab-completion** plugin.

~~I have a plugin **git-flow-completion** how to make **fzf-tab** turn off when I use **git**?~~
The second question partially resolved **->** added a function in the plugin and hung in `.zshrc` 
`bindkey '^ T' fzf-tab-toggle` and sent a [pull request](https://github.com/Aloxaf/fzf-tab/pull/30).
Of course, I would like it to be done automatically.

I have **fzf-tab-completion** plugin does not work on the **TAB** key.
But `bindkey '^I' fzf_completion` is not convenient. 
Therefore, I prefer the **fzf-tab** plugin.

This is displayed in **fzf-tab** which is not very convenient:
here the keys are scattered

This is displayed in **fzf-tab-completion** which is very convenient:
in the example, the keys **-A --all --no-ignore-removal** are located together

Aloxaf:
Fix #29

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -53,7 +53,7 @@ compadd() {
 
     # dscr - the string to show to users
     # word - the string to be inserted
-    local dscr word i
+    local dscr word i cnt=$#_fzf_tab_compcap
     for i in {1..$#__hits}; do
         word=$__hits[i] && dscr=$__dscr[i]
         if [[ -n $dscr ]]; then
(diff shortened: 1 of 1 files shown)


Aloxaf:
Close due to #46

PR #36, opened 2020-02-17
bug fix: don't break user locale

romkatv:
There is a bug in Zsh that causes `local LC_ALL=C` to leak outside of its scope. As a result, in some environments fzf-tab changes user locale.

It's not important for candidates to be sorted in byte order, and it's also arguably worse than respecting the order prescribed by the user locale. So I'm removing it.

I think it was I who added this `local LC_ALL=C` in an earlier PR. Sorry about that.

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -232,7 +232,6 @@ _fzf_tab_get_candidates() {
     (( $#candidates == 0 )) && return
 
     (( same_word )) && candidates[2,-1]=()
-    local LC_ALL=C
     candidates=("${(@on)candidates}")
 
     # hide needless group


Aloxaf:
In fact, it's me who added `LC_ALL=C` at first.

I added this because `sort` will always ignore `-` if I don't set `LC_ALL=C`
image

Seems also related to #29

Aloxaf:
Could you tell me in which situation will `local LC_ALL=C` leak outside of its scope?

romkatv:
> Could you tell me in which situation will `local LC_ALL=C` leak outside of its scope?

```zsh
(
  unset -m 'LC_*|LANG'
  export LC_CTYPE='en_US.UTF-8'  # set this to any UTF-8 locale you have
  echo '\u276F'  # this works
  () {
    local LC_ALL=C
  }
  echo '\u276F'  # this doesn't work
)
```

Since this code no longer calls `sort`, the original reason for setting `LC_ALL=C` is gone, right?

Aloxaf:
zsh's builtin array sort also respects to this variable.

Maybe this will be better?
```zsh
(
  unset -m 'LC_*|LANG'
  export LC_CTYPE='en_US.UTF-8'  # set this to any UTF-8 locale you have
  echo '\u276F'  # this works
  () {
    local LC_ALL= LC_COLLATE=C
    print -l -- -a --a --b --c  -b -c | sort  # output is correct
  }
  echo '\u276F'  # and this works
)
```

romkatv:
It breaks this though:

```zsh
(
  unset -m 'LC_*|LANG'
  LC_COLLATE=en_US.UTF-8
  x=(-a --b)
  print -r -- ${(on)x}  # this prints "-a --b"
  () {
    local LC_ALL= LC_COLLATE=C
    print -r -- ${(on)x}
  }
  print -r -- ${(on)x}  # this prints "--b -a"
)
```

> zsh's builtin array sort also respects to this variable.

Maybe respecting the user-defined `LC_COLLATE` is the right thing to do here?

Aloxaf:
You're right. Thanks a lot.

PR #41, opened 2020-02-21
Support FZF custom completion API

doubleloop:
API explained [here](https://github.com/junegunn/fzf/wiki/Examples-(completion)), handy way of defining custom completions for particular commands.
Copied some functions from fzf [`completions.zsh`](https://github.com/junegunn/fzf/blob/master/shell/completion.zsh) to allow not sourcing it.

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -85,6 +85,8 @@ _fzf_tab_remove_space() {
 : ${FZF_TAB_SHOW_GROUP:=full}
 : ${FZF_TAB_NO_GROUP_COLOR:=$'\033[37m'}
 : ${FZF_TAB_CONTINUOUS_TRIGGER:='/'}
+: ${FZF_TAB_CUSTOM_COMPLETIONS:='1'}
+: ${FZF_TAB_CUSTOM_COMPLETIONS_PREFIX:='_fzf_complete_'}
 : ${(A)=FZF_TAB_QUERY=prefix input first}
 : ${(A)=FZF_TAB_SINGLE_GROUP=color header}
 : ${(A)=FZF_TAB_GROUP_COLORS=\
@@ -95,7 +97,7 @@ _fzf_tab_remove_space() {
 
 (( $+FZF_TAB_OPTS )) || FZF_TAB_OPTS=(
     --ansi   # Enable ANSI color support, necessary for showing groups
-    --expect='$FZF_TAB_CONTINUOUS_TRIGGER' # For continuous completion 
+    --expect='$FZF_TAB_CONTINUOUS_TRIGGER' # For continuous completion
     '--color=hl:$(( $#headers == 0 ? 108 : 255 ))'
     --nth=2,3 --delimiter='\x00'  # Don't search FZF_TAB_PREFIX
     --layout=reverse --height=75%
@@ -306,7 +308,24 @@ _fzf_tab_complete() {
 
 zle -C _fzf_tab_complete complete-word _fzf_tab_complete
 
+_fzf_tab_try_custom_completion() {
+    # do not steal fzf's completions
+    [[ $LBUFFER =~ ${(q)FZF_COMPLETION_TRIGGER-'**'}$ ]] && return 1
+    local tokens=(${(z)LBUFFER})
+    [[ ${LBUFFER[-1]} = ' ' ]] && tokens+=("")
+    local cmd=${tokens[1]}
+    if (( $+functions[${FZF_TAB_CUSTOM_COMPLETIONS_PREFIX}${cmd}] )); then
+        local prefix=${tokens[-1]}
+        local lbuf
+        [ -z "${tokens[-1]}" ] && lbuf=$LBUFFER || lbuf=${LBUFFER:0:-${#tokens[-1]}}
+        prefix="$prefix" eval _fzf_complete_${cmd} ${(q)lbuf}
+        return 0
+    fi
+    return 1
+}
+
 fzf-tab-complete() {
+    (( FZF_TAB_CUSTOM_COMPLETIONS )) && _fzf_tab_try_custom_completion && return
     # complete or not complete, this is a question
     # this name must be ugly to avoid clashes
     local -i _fzf_tab_continue=1 _fzf_tab_should_complete=0
--- a/README.md
+++ b/README.md
@@ -59,6 +59,25 @@ Key Bindings:
 
 For example <kbd>Ctrl</kdb>+<kdb>T</kbd> `bindkey '^T' toggle-fzf-tab`
 
+### Custom completions
+
+There exists mechanism for overwriting completion action for particular command
+[similar to fzf](https://github.com/junegunn/fzf/wiki/Examples-(completion)).
+If function named `_fzf_complete_foo` is found, it will be used for handling completions
+of `foo` command.
+
+If you also have [fzf's completions](https://github.com/junegunn/fzf#fuzzy-completion-for-bash-and-zsh)
+enabled (`completion.zsh` is sourced), you can use it's `_fzf_complete` helper function, for example:
+```
+_fzf_complete_foo() {
+  _fzf_complete "--multi --reverse" "$@" < <(
+    echo foo
+    echo bar
+    echo bazz
+  )
+}
+```
+
 ### Configure
 
 Here are some variables which can be used to control the behavior of fzf-tab.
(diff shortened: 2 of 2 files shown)


fzf-tab.zsh
```diff
@@ -306,7 +306,50 @@ _fzf_tab_complete() {
 
 zle -C _fzf_tab_complete complete-word _fzf_tab_complete
 
+__fzfcmd_complete() {
+  [ -n "$TMUX_PANE" ] && [ "${FZF_TMUX:-0}" != 0 ] && [ ${LINES:-40} -gt 15 ] &&
+    echo "fzf-tmux -d${FZF_TMUX_HEIGHT:-40%}" || echo "fzf"
+}
+
+_fzf_feed_fifo() (
```
romkatv: If you copy functions over, give them a different name. Otherwise, when these functions get modified in fzf, the behavior of fzf-tab, and (worse!) of plain fzf will depend on the order in which files are sourced. It would be quite bad if fzf-tab affects the behavior of plan fzf.

Aloxaf:
Thanks for your contribution.

But fzf-tab should be able to work together with fzf's completions.zsh, and its API is experimental. What's the advantage of supporting it?

doubleloop:
> But fzf-tab should be able to work together with fzf's completions.zsh

I will try to fix this

>What's the advantage of supporting it

The ability to customize for particular commands is really cool. Actually It does not have to be fzf API, could be anything thet allow user to use zsh sources by default but provide convenient way to define custom behavior for for particular commands. Still, FZF mechanism for customization is quite convenient and easy to use and if user already defined some custom `_fzf_complete_<cmd>` functions, it could be a benefit if FZF-tab recognize and use them  (there could be some env variable like `FZF_TAB_USE_FZF_CUSTOM_COMPLETIONS`) .

romkatv:
> The ability to customize for particular commands is really cool.

Doesn't Zsh have this feature out of the box? `compdef _pids foo` will complete command `foo` using completion function `_pids`. Instead of `_pids` you can use your own completion function.

doubleloop:
> Doesn't Zsh have this feature out of the box?

True. On the other hand, providing some framework/helper function like `_fzf_complete` out of the box make customizations much easier to write.

Consider for example:
```zsh
BLUE='\033[0;34m'
RED='\033[0;31m'
NC='\033[0m'

_foo() {
    for i in {1..40}; do
        echo -e "${BLUE}foo${NC} completion$i"
    done
}

_bar() {
    # this is some slower command
    sleep 1
    for i in {40..80}; do
        echo -e "${RED}bar${NC} completion$i"
    done
}

_fzf_complete_foobar() {
    _fzf_complete "--multi --ansi -n 2" "$@" < <(
        _foo
        _bar
    )
}
_fzf_complete_foobar_post() { awk '{print $2}'; }
```

How to easily write custom `compdef` function which:
1. is asynchronous (does not delay showing completions untill bar is finished)
1. is easy to specify if --multi vs --no-multi should be used (`rm` vs `cd`)
1. is easy to provide some coloring/icons
1. specify which part of sorce line should be used for fuzzy search (`-n` flag)

I do not know zsh completion mechanism well, so after few minutes of reading some docs I started with something like:
```
_foobar() {
    completions=(${(f)"$(_foo)"})
    compadd $completions
    completions=(${(f)"$(_bar)"})
    compadd $completions
}
compdef _foobar foobar
```
From what I understand, using compdef mechanism it is hard to controll multi/no-multi and coloring, also completion is synchronous.

romkatv:
I see. Streaming completions would definitely be useful.

doubleloop:
>> But fzf-tab should be able to work together with fzf's completions.zsh
>
> I will try to fix this

works together wit fzf completions now

To enable, set `FZF_TAB_USE_FZF_CUSTOM_COMPLETIONS=1`

Aloxaf:
> I will try to fix this

Sorry I didn't make myself clear.
I mean, if I can get the support of fzf's custom completion API by sourcing [completion.zsh](https://github.com/junegunn/fzf/blob/master/shell/completion.zsh), what's the advantage of supporting it in fzf-tab itself.

doubleloop:
> if I can get the support of fzf's custom completion API by sourcing completion.zsh, what's the advantage of supporting it in fzf-tab itself.

No need for separate completion trigger. Currently, when one define some custom completion,  one has to explicitly use fzf trigger (`**<tab>`) to delegate it to `completion.zsh` while fzf-tab (`<tab>`) for everything else. One can change `FZF_COMPLETION_TRIGGER` or change the shortcut (described [here](https://github.com/junegunn/fzf/wiki/Configuring-fuzzy-completion#dedicated-completion-key)), but still in result there are 2 separate triggers, which is (arguably) inconvenient. With this change one can use `<tab>`, so that if custom completion function is defined for particular command, it is used, otherwise fzf-tab completion is used. One still cen explicitly delegate to `completion.zsh` using `FZF_COMPLETION_TRIGGER`.

Aloxaf:
Sounds great. Please update README.md as well.

doubleloop:
I cleaned up the PR and updated `README.md`.

After some thought, I decided that the best solution is not providing `_fzf_complete` helper function. Ability to hook custom functions is enough. How they will be used is up to user.

fzf-tab.zsh
```diff
@@ -306,7 +308,24 @@ _fzf_tab_complete() {
...
+        local lbuf
+        [ -z "${tokens[-1]}" ] && lbuf=$LBUFFER || lbuf=${LBUFFER:0:-${#tokens[-1]}}
+        prefix="$prefix" eval _fzf_complete_${cmd} ${(q)lbuf}
+        return 0
+    fi
+    return 1
+}
+
 fzf-tab-complete() {
+    [[ -n $FZF_TAB_CUSTOM_COMPLETIONS ]] && _fzf_tab_try_custom_completion && return
```
Aloxaf: Should it be `(( $FZF_TAB_CUSTOM_COMPLETIONS ))` or `[[ $FZF_TAB_CUSTOM_COMPLETIONS == 1 ]]` here?

Or can we just check if `$FZF_TAB_CUSTOM_COMPLETIONS` exists? ( set to non-null value to enable, set to empty value to disable )

doubleloop: should be (( $FZF_TAB_CUSTOM_COMPLETIONS ))

romkatv: It must not have the dollar sign. This is correct: `(( FZF_TAB_CUSTOM_COMPLETIONS ))`. With the dollar sign you are also supporting indirection that you don't want, and potentially even executing arbitrary code.

doubleloop: changed

fzf-tab.zsh
```diff
@@ -306,7 +308,24 @@ _fzf_tab_complete() {
 
 zle -C _fzf_tab_complete complete-word _fzf_tab_complete
 
+_fzf_tab_try_custom_completion() {
+    # do not steal fzf's completions
+    [[ $LBUFFER =~ ${(q)FZF_COMPLETION_TRIGGER-'**'}$ ]] && return 1
```
romkatv: I set `FZF_COMPLETION_TRIGGER=""` and bind `fzf-completion` to a separate key (Ctrl-T). This way I can use TAB for normal completions and Ctrl-T to complete file arguments with fzf. I find that pressing Ctrl-T to complete file arguments with fzf is more convenient than pressing \*-\*-TAB.

With this setup `_fzf_tab_try_custom_completion` always bails out early.

Does the check on the first line of `_fzf_tab_try_custom_completion` presuppose that `_fzf_tab_orig_widget` is `fzf-completion` (or some other widget that respects `FZF_COMPLETION_TRIGGER`)? I set `_fzf_tab_orig_widget` to `expand-or-complete`, so this check doesn't make much sense.

FWIW, I don't know what's the right thing to do here. I'm unsure how exactly I want completions to work. Just wanted to share that the current code is a bit odd and likely incorrect.

doubleloop: I just checked for this configuration:
```
export FZF_COMPLETION_TRIGGER=""
bindkey '^T' fzf-completion
bindkey '^I' fzf-tab-complete

export fzf_default_completion=expand-or-complete
export _fzf_tab_orig_widget=expand-or-complete
```

If no custom completion function is defined, TAB triggers fzf-tab, ctrl-t triggers fzf.
If some custom completion function is defined, both TAB and ctrl-t triggers it.

>_fzf_tab_try_custom_completion always bails out early
You mean it return 1 on the first if?
```
$ FZF_COMPLETION_TRIGGER=""
$ echo ${(q)FZF_COMPLETION_TRIGGER-'**'}
''
```
so if I am correct, everything should match..

romkatv: > You mean it return 1 on the first if?

Yes.

> If some custom completion function is defined, both TAB and ctrl-t triggers it.

Are you sure TAB triggers it? It doesn't for me. When I press TAB, `_fzf_tab_try_custom_completion` returns on the first line, so `_fzf_complete_*` is never called. Can you verify that `_fzf_tab_orig_widget` is indeed `expand-or-complete` in your setup?

PR #46, opened 2020-02-28
More configuration via zstyle

### LINKED ISSUE #29: Show identical keys nearby

Aloxaf:
Fix #29 

Also provide many other features.

```zsh
# disable sort when completing options of any command
zstyle ':completion:complete:*:options' sort false

# use input as query string when completing zlua
zstyle ':fzf-tab:complete:_zlua:*' query-string input

# give a preview when completing `kill`
zstyle ':completion:*:*:*:*:processes' command "ps -u $USER -o pid,user,comm,cmd -w -w"
zstyle ':fzf-tab:complete:kill:argument-rest' extra-opts '--preview=echo $(<{f})' --preview-window=down:3:wrap

# (experimental) give a preview of directory when compleing cd
local extract="
# trim input
in=\${\${\"\$(<{f})\"%\$'\0'*}#*\$'\0'}
# get ctxt for current completion
local -A ctxt=(\"\${(@ps:\2:)CTXT}\")
"
zstyle ':fzf-tab:complete:cd*' extra-opts --preview=$extract"exa -1 --color=always \${~ctxt[hpre]}\$in"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@ Replace zsh's default completion selection menu with fzf!
         - [single-group](#single-group)
         - [group-colors](#group-colors)
         - [show-group](#show-group)
-        - [sort](#sort)
 - [Difference from other plugins](#difference-from-other-plugins)
 - [Compatibility with other plugins](#compatibility-with-other-plugins)
 - [Related projects](#related-projects)
@@ -95,7 +94,7 @@ Now it use zstyle, because zstyle can give you more control over fzf-tab's behav
 
 ```zsh
 # disable sort when completing options of any command
-zstyle ':fzf-tab:complete:*:options' sort false
+zstyle ':completion:complete:*:options' sort false
 
 # use input as query string when completing zlua
 zstyle ':fzf-tab:complete:_zlua:*' query-string input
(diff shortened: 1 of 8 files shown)


romkatv:
By the way, there is `sort` style in the standard completion system. Search for "This allows the standard ordering of matches to be overridden" in http://zsh.sourceforge.net/Doc/Release/Completion-System.html.

Perhaps fzf-tab shouldn't be sorting given that users can configure this in the underlying completion system with more flexibility? What do you think?

Aloxaf:
But this sort seems to be done after `compadd`, in other words, fzf-tab still has to decide whether to sort the results by itself. ( But it can use the default style instead of introducing a new one )

romkatv:
> But this sort seems to be done after `compadd`

I didn't realize this. This is obvious in retrospect.

> But it can use the default style instead of introducing a new one

That's a good idea.

PR #48, opened 2020-02-28
fix a few bugs in _fzf_tab_try_custom_completion

romkatv:
- make sure all sublists succeed when `err_return` might be set
- don't assume that `${(q)...}` correctly escapes regex (it doesn't)
- don't use `${x[1]}` when `ksh_arrays` might be set (it makes arrays 0-based)
- don't hard-code `_fzf_complete_`
- don't eval the first token in `$LBUFFER`
- don't export `prefix`

@doubleloop Could you double-check that this change doesn't break stuff?

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -308,24 +308,29 @@ _fzf_tab_complete() {
 
 zle -C _fzf_tab_complete complete-word _fzf_tab_complete
 
+# note: this function is called with user options
 _fzf_tab_try_custom_completion() {
     # do not steal fzf's completions
-    [[ $LBUFFER =~ ${(q)FZF_COMPLETION_TRIGGER-'**'}$ ]] && return 1
-    local tokens=(${(z)LBUFFER})
-    [[ ${LBUFFER[-1]} = ' ' ]] && tokens+=("")
-    local cmd=${tokens[1]}
-    if (( $+functions[${FZF_TAB_CUSTOM_COMPLETIONS_PREFIX}${cmd}] )); then
-        local prefix=${tokens[-1]}
-        local lbuf
-        [ -z "${tokens[-1]}" ] && lbuf=$LBUFFER || lbuf=${LBUFFER:0:-${#tokens[-1]}}
-        prefix="$prefix" eval _fzf_complete_${cmd} ${(q)lbuf}
-        return 0
-    fi
-    return 1
+    [[ $LBUFFER != *"${FZF_COMPLETION_TRIGGER-**}" ]] || return
+    local func prefix lbuf
+    () {
+        emulate -L zsh
+        local tokens=(${(z)LBUFFER})
+        func=${FZF_TAB_CUSTOM_COMPLETIONS_PREFIX}${tokens[1]}
+        (( $+functions[$func] )) || return
+        if [[ $LBUFFER == *' ' ]]; then
+          lbuf=$LBUFFER
+        else
+          prefix=$tokens[-1]
+          lbuf=${LBUFFER:0:-$#prefix}
+        fi
+    } || return
+    # must run with user options; don't add `emulate -L zsh` above this line
+    "$func" "$lbuf" || true
 }
 
 fzf-tab-complete() {
-    (( FZF_TAB_CUSTOM_COMPLETIONS )) && _fzf_tab_try_custom_completion && return
+    (( ! FZF_TAB_CUSTOM_COMPLETIONS )) || ! _fzf_tab_try_custom_completion || return 0
     # complete or not complete, this is a question
     # this name must be ugly to avoid clashes
     local -i _fzf_tab_continue=1 _fzf_tab_should_complete=0


fzf-tab.zsh
```diff
@@ -308,24 +308,29 @@ _fzf_tab_complete() {
...
+        (( $+functions[$func]} )) || return
+        if [[ $LBUFFER == *' ' ]]; then
+          lbuf=$LBUFFER
+        else
+          prefix=$tokens[-1]
+          lbuf=${LBUFFER:0:-$#prefix}
+        fi
+    } || return
+    # must run with user options; don't add `emulate -L zsh` above this line
+    prefix=$prefix "$func" "$lbuf" || true
```
romkatv: Quick question: does `prefix` really need to be exported? It's quite unusual to export lower-case variables. If I don't export it, all `_fzf_complete_*` functions still seem to work.

romkatv: I've removed this `export`. @doubleloop, please give a holler if it's really necessary to export `prefix`.

doubleloop: it is used in [_fzf_complete](https://github.com/junegunn/fzf/blob/master/shell/completion.zsh#L119) as `-q` parameter, this is why I exported it

romkatv: Thanks for the context. Since the code you linked to doesn't care whether this variable is exported or not, it's safe to not export it, right? The fact that `prefix` is passed as `-q` command line argument to `fzf` strongly suggests that it's unnecessary to export it.

doubleloop: Ok, seems I do not know zsh well.. I supposed:
```zsh
$ foo() { local xxx=1; bar }
$ bar(){ echo $xxx }
$ foo
```
will not print anything because `xxx` is not visible in bar. So is there any particular reason why fzf exports it [here](https://github.com/junegunn/fzf/blob/master/shell/completion.zsh#L202)?

romkatv: > I supposed...

Thanks for clarifying this! It clears up the confusion we had. Dynamically scoped parameters come from POSIX, so all POSIX-based shells behaves like Zsh in this regard.

> So is there any particular reason why fzf exports it here?

Looks like a bug. That code has many bugs.

fzf-tab.zsh
```diff
@@ -308,24 +308,29 @@ _fzf_tab_complete() {
...
-    local cmd=${tokens[1]}
-    if (( $+functions[${FZF_TAB_CUSTOM_COMPLETIONS_PREFIX}${cmd}] )); then
-        local prefix=${tokens[-1]}
-        local lbuf
-        [ -z "${tokens[-1]}" ] && lbuf=$LBUFFER || lbuf=${LBUFFER:0:-${#tokens[-1]}}
-        prefix="$prefix" eval _fzf_complete_${cmd} ${(q)lbuf}
-        return 0
-    fi
-    return 1
+    [[ $LBUFFER != *"${FZF_COMPLETION_TRIGGER-**}" ]] || return
```
doubleloop: `_fzf_tab_orig_widget` should be checked (if it is not fzf-completion, checking FZF_COMPLETION_TRIGGER is pointless, previously this worked because of `q` bug).
```
[[ $_fzf_tab_orig_widget != 'fzf-completion' || $LBUFFER != *"${FZF_COMPLETION_TRIGGER-**}" ]] || return
```

You can quickly check completions of foo in this settings
```
export FZF_COMPLETION_TRIGGER=""
bindkey '^T' fzf-completion
bindkey '^I' fzf-tab-complete
export fzf_default_completion=expand-or-complete
export _fzf_tab_orig_widget=expand-or-complete
_fzf_complete_foo() { _fzf_complete "+m -n 1" "$@" < <(echo bar; echo baz) }
```

romkatv: What you want to check for is whether `zle $_fzf_tab_orig_widget` recognizes `FZF_COMPLETION_TRIGGER`. There is no way to implement this check in the code. Zle widgets get renamed and wrapped left and right. A widget called `expand-or-complete` can very well respect `FZF_COMPLETION_TRIGGER`, while a widget called `fzf-completion` may ignore it.

I'm having mixed feelings about the introduction of custom completions to fzf-tab. fzf-tab used to be focused on doing one thing well -- using fzf to display completion candidates from from the builtin Zsh completion system. Now it also has its own system for generating completion candidates. This poses many problems:

1. How to decide which system to use to generate completions? This is what has started the discussion of `FZF_COMPLETION_TRIGGER`.

2. The alternative completion system has some advantages (e.g., completion candidates can be streamed into fzf instead of accumulating all candidates before firing up fzf) but it also has many disadvantages. The biggest disadvantages is that it doesn't understand zsh syntax.

Users who happened to be using fzf-tab and fzf will now find that completions for several commands (`export`, `unset`, `ssh`, etc.) don't work as well as they used to. For example:

```zsh
unset foo; cd ./<TAB>
```

This used to work. Now it doesn't because it'll trigger custom completion for `unset`.

Perhaps fzf-tab shouldn't provide this extension mechanism until we better understand how it should work? Users who want it can implement their own dispatching widget and bind it to TAB. The dispatching widget can call `fzf-tab-complete`, `fzf-complete` or whatever, based on conditions preferred by the user.

WDYT?

doubleloop: > unset foo; cd ./<TAB>

true, did not consider all cases.

>Perhaps fzf-tab shouldn't provide this extension mechanism until we better understand how it should work?

No problem with me.

romkatv: @Aloxaf I propose to merge this PR as it makes code strictly better by fixing a few bugs, and then revert custom completions. Hopefully there aren't many users who rely on this feature yet (just one?) and they can have it by adding a few lines of code to their own configs. These lines can be based on the code from fzf-tab that is going to be reverted, so it would be nice if bug fixes from this PR landed first.

fzf-tab.zsh
```diff
@@ -308,24 +308,29 @@ _fzf_tab_complete() {
...
+    local func prefix lbuf
+    () {
+        emulate -L zsh
```
doubleloop: Can I ask what is the benefit of `emulate -L zsh` in cases like this?

romkatv: The behavior of zsh can be changed with options. For example, `setopt ksh_arrays` has the following effects:

1. Arrays and scalars are now 0-based.
2. `$#foo` and `${#foo}` expand to the length of the first element in the array instead of the length of the array itself.
3. A few other things.

It's extremely difficult to write code that behaves correctly with all possible options, which on the surface would make it impossible to define functions that everyone could use regardless of their preferred options. To solve this problem zsh has builtin `emulate` that allows you to set all options to their default values. You can use `emulate sh` to make zsh behave like `sh` (some users actually put that in their `zshrc`), `emulate zsh` to make it behave "in native mode", and a few other modes. However, if you put `emulate zsh` in a function, you'll override options the user chose for themselves. Perhaps they put `setopt extended_glob` in `zshrc` but after calling the function this option is gone. To solve *this* problem there is `local_options` option, which causes options to be restored when function returns. `emulate -L zsh` is like `emulate zsh` plus `local_options` (and a few other `local_*` things).

Writing zsh functions that work for everyone is *really difficult*. In comparison, writing zsh scripts is relatively straightforward.

fzf-tab.zsh
Aloxaf: No problem. Thanks for both of your impressive discussions.

PR #49, opened 2020-02-29
Add toggle sort in plugin

webmastak:
Add toggle sort in plugin, <kbd>Ctrl</kdb>+<kdb>T</kbd> `bindkey '^T' toggle-fzf-tab`

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -32,6 +32,9 @@ compadd() {
         return
     fi
 
+    # store $curcontext for furthur usage
+    _fzf_tab_curcontext=:fzf_tab:${curcontext}
+
     # keep order of group description
     [[ -n $expl ]] && _fzf_tab_groups+=$expl
 
@@ -53,7 +56,8 @@ compadd() {
 
     # dscr - the string to show to users
     # word - the string to be inserted
-    local dscr word i
+    local dscr word i sort cnt=$#_fzf_tab_compcap
+    zstyle -b "$_fzf_tab_curcontext" sort sort
     for i in {1..$#__hits}; do
         word=$__hits[i] && dscr=$__dscr[i]
         if [[ -n $dscr ]]; then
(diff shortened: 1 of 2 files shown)


romkatv:
Did you just copy changes from https://github.com/Aloxaf/fzf-tab/pull/46?

fzf-tab.zsh
```diff
@@ -358,15 +380,36 @@ enable-fzf-tab() {
...
+    fi
+}
+
+disable-sort-fzf-tab() {
+    typeset -g _ZSH_FZF_TAB_SORT
+    emulate -L zsh
+	zstyle ':fzf_tab:*' sort false 
+}
+
+enable-sort-fzf-tab() {
```
romkatv: Adding enable/disable functions for every boolean zstyle options is not wise.

webmastak: However, it works and not bad.

romkatv: It is bad as it creates unnecessary API bloat. Every public symbol has to be documented, read by users and supported forever.

webmastak: Can you make the sorting correct?

Aloxaf:
1. Don't copy. You can use `git checkout -b add-toogle-sort configure_by_zstyle` to create a new branch based on #46, which is still under develop and may change in the future.
2. I agree with romkatv. `toggle-sort-fzf-tab` is not a necessary feature, it won't be frequently used by most users and it can be implemented easily by users themselves.

fzf-tab.zsh
Aloxaf: Does `zstyle ':fzf_tab:*:options' sort false` work as your expect?
图片

webmastak:
I will leave her for myself.

fzf-tab.zsh
Aloxaf: Oh I introduced a bug again, fixed in https://github.com/Aloxaf/fzf-tab/pull/46/commits/d2a8074f63a621e13b7e4f525c746448be71d589

Aloxaf: Oh I introduced a bug again, fixed in https://github.com/Aloxaf/fzf-tab/pull/46/commits/d2a8074f63a621e13b7e4f525c746448be71d589.
I need to write some tests...

PR #50, opened 2020-03-06
minor and not-so-minor cleanup of the recent changes

romkatv:
- Avoid running code with unknown options whenever possible. Put initialization code in an anonymous function with `emulate -L zsh`.
- Use only two sets of options for all code: (1) user-defined options only when it's strictly necessary; (2) the same set of "good" options everywhere else. This makes it much easier to reason about code.
- Remove `_ZSH_FZF_TAB_DISABLED`. It's inconsistently named, awkwardly used and is unnecessary.
- Don't rely on undocumented implementation details of zsh-autosuggestions.

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -102,32 +102,38 @@ _fzf_tab_remove_space() {
     '--header-lines=$#headers' # $#headers will be expanded to lines of headers at runtime
 )
 
-_fzf_tab_add_default() {
-    zstyle -t ':fzf-tab:*' $1
-    (( $? != 2 )) || zstyle ':fzf-tab:*' $1 ${@:2}
-}
-
 _fzf_tab_get() {
     zstyle $1 ":fzf-tab:$_fzf_tab_curcontext" ${@:2}
 }
 
-# Some users may still use variable
-_fzf_tab_add_default continuous-trigger ${FZF_TAB_CONTINUOUS_TRIGGER:-'/'}
-_fzf_tab_add_default fake-compadd ${FZF_TAB_FAKE_COMPADD:-default}
-_fzf_tab_add_default insert-space ${FZF_TAB_INSERT_SPACE:-true}
-_fzf_tab_add_default query-string ${(A)=FZF_TAB_QUERY:-prefix input first}
-_fzf_tab_add_default single-group ${(A)=FZF_TAB_SINGLE_GROUP:-color header}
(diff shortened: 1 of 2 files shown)


fzf-tab.zsh
```diff
@@ -374,24 +374,31 @@ fzf-tab-complete() {
...
-    unset _ZSH_FZF_TAB_DISABLED
-    emulate -L zsh
-    typeset -g _fzf_tab_orig_widget="${$(bindkey '^I')##* }"
+    emulate -L zsh -o extended_glob
+    (( $+_fzf_tab_orig_widget )) && return
+
+    typeset -g _fzf_tab_orig_widget="${${$(bindkey '^I')##* }:-expand-or-complete}"
+    if (( ! $+widgets[.fzf-tab-orig-$_fzf_tab_orig_widget] )); then
+      # Copy the widget before it's wrapped by zsh-autosuggestions and zsh-syntax-highlighting.
+      # Prefix the name of the copy with '.' so that it doesn't get wrapped.
```
Aloxaf: This will also prevent `compinit` from editing the widget, which will make fzf-tab unusable if it initializes before call `compinit`.

romkatv: Good point!

How about now?

fzf-tab.zsh
```diff
@@ -374,24 +374,49 @@ fzf-tab-complete() {
...
+            expand-or-complete
+            expand-or-complete-prefix
+            list-choices
+            menu-complete
+            menu-expand-or-complete
+            reverse-menu-complete
+        )
+        # Note: We prefix the name of the widget with '.' so that it doesn't get wrapped.
+        if [[ $widgets[$_fzf_tab_orig_widget] == builtin &&
+              $compinit_widgets[(Ie)$_fzf_tab_orig_widget] == 1 ]]; then
```
Aloxaf: I think here should be `!=0`
```suggestion
              $compinit_widgets[(Ie)$_fzf_tab_orig_widget] != 0 ]]; then
```

romkatv: Thanks! Applied your suggested change.

romkatv:
Fixed tests by removing `_fzf_tab_orig_widget=expand-or-complete` from `test/fzftab.ztst`.

fzf-tab.zsh
Aloxaf: LGTM

Aloxaf:
LGTM. Thanks a lot!

romkatv:
Thanks for the quick and thorough review!

PR #54, opened 2020-03-09
Call _fzf_tab_complete directly in _main_complete

### LINKED ISSUE #47: Warning message is missing if load zsh-autosuggestions
```zsh
zstyle ':completion:*:warnings' format '%F{red}%B-- No match for: %d --%b%f'
```

This bug was introduced by e18057e2d81bca4643eeef3dcc828f90b662952f

### LINKED ISSUE #53: Conflict with zdharma/fast-syntax-highlighting
fzf-tab doesn't trigger properly on `<tab>` when loaded after the [fast-syntax-highlighting](/zdharma/fast-syntax-highlighting) plugin. Prior to b856498 the plugin worked when loading in either order. 

I traced the issue to the following subset of the change: https://github.com/Aloxaf/fzf-tab/commit/b856498fe21642af5fc33154b39a5b18999759c4#diff-3de94c7df288cd22645487b69fdd019cL358-R363

Reverting that fixes the issue, but I do not completely understand why it was changed in the first place or how to fix it.

Aloxaf:
Fix #47

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -283,7 +283,7 @@ _fzf_tab_complete() {
     local choice choices _fzf_tab_curcontext continuous_trigger
 
     IN_FZF_TAB=1
-    _main_complete  # must run with user options; don't move `emulate -L zsh` above this line
+    orig_main_complete  # must run with user options; don't move `emulate -L zsh` above this line
     IN_FZF_TAB=0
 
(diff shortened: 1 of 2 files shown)


Aloxaf:
This can only fix #47. And `comppostfuncs` seems to be a better choice.

PR #57, opened 2020-03-10
Add support of _approximate completer

Aloxaf:


Aloxaf:
I don't know why the result of `_approximate` can't be inserted. This may take a long time, I will merge some useful changes to master.

alberti42:
I wasn't aware of the 2020 attempt on the `approximate` branch - I just found out about it. I went back and skimmed the code to understand the direction it took.

My current PR https://github.com/Aloxaf/fzf-tab/pull/554 follows the same general idea (keeping `_approximate` inside fzf-tab), but it also addresses additional edge cases that were still causing the completion UI to fall back or insert the wrong prefix. It now passes the existing test suite.

I'd appreciate it if you could try the new implementation. Your feedback would be especially helpful to catch any remaining edge cases.

#### TL;DR:

What the old approximate branch did (high level)

- It monkey-patched _approximate by copying its source and replacing builtin compadd with _fzf_tab_compadd, so fzf-tab could capture matches (fzf-tab.zsh:579).
- It wrapped _approximate and temporarily removed the global compadd function while _approximate ran, so Zsh would define its local wrapper (fzf-tab.zsh:580-584).
- It restored compadd and _approximate when disabling fzf-tab (fzf-tab.zsh:520-525).

That approach mostly targeted the "compadd bypass" problem. It didn't handle the other conflicts that make _approximate fall back to the standard menu or insert wrong prefixes.

How the current PR differs (high level)

- It still patches _approximate, but does it in a more controlled wrapper (restores hooks in always, and forces IN_FZF_TAB so the capture path is reliable).
- It sanitizes PREFIX when _approximate injects (#aN) so captured matches align with what Zsh expects to insert.
- It respects compstate[list]=force so fzf-tab doesn't auto-accept an unambiguous prefix and skip the list.
- It applies choices even when the original widget returns non-zero (common in _approximate/_correct list-only flows).
- It passes the existing test suite.

PR #70, opened 2020-04-03
replace `sort -t '\0'` with a portable alternative

romkatv:
`sort -t '\0'` doesn't work on busybox:

```text
❯ sort -t '\0'
sort: bad -t parameter
❯ echo ${commands[sort]:A}
/bin/busybox
```

I've replaced `sort -t '\0' -k2` with this sequence:

1. move the first field to the end
2. sort
3. move the last field to the front

I've made a couple additional loosely related changes that seemed to make sense. Please double-check that my reasoning is correct.

1. I've removed number sorting: `-n` in `sort` and `(n)` in zsh. This code is sorting either completion candidates themselves or their group names. Neither are numbers. I don't think it makes much of a difference either way, so if there is even a slight reason to use number sorting, I won't mind it in the slightest.
2. The original sort had `-u` which doesn't seem necessary. I've removed it.

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -251,7 +251,7 @@ _fzf_tab_colorize() {
 # pupulates array `candidates` with completion candidates
 _fzf_tab_get_candidates() {
     local dsuf dpre k _v filepath first_word show_group no_group_color prefix bs=$'\b'
-    local -a list_colors group_colors tcandidates
+    local -a list_colors group_colors tcandidates match mbegin mend
     local -i  same_word=1 colorful=0
     local -Ua duplicate_groups=()
     local -A word_map=()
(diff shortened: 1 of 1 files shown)


Aloxaf:
Sorry I was busy with Tomb Sweeping festival last week.

This seems to be nice.

romkatv:
> Sorry I was busy with Tomb Sweeping festival last week.

Welcome back!

> This seems to be nice.

Could you merge it then?

Aloxaf:
I find there is a noticeable performance loss on my machine. How about making the delimiter configurable?

```zsh
repeat 5000 tcandidates+=($'abc\0def')
time (
  _tcandidates=(${(@o)${(@)tcandidates:/(#b)([^$'\0']#)$'\0'(*)/$match[2]$'\0'$match[1]}})
  _tcandidates=(${(@)_tcandidates/(#b)(*)$'\0'([^$'\0']#)/$match[2]$'\0'$match[1]})
)
time (_tcandidates=(${(f)"$(sort -u -t '\0' -k 2 <<< ${(pj:\n:)tcandidates})"}))
```

```zsh
( _tcandidates=() ; _tcandidates=() ; )
 User: 0.05s    System: 0.00s  Total: 0.047s
 CPU:  98%      Mem:    12 MB
( _tcandidates=(${(f)"$(sort -u -t '\0' -k 2 <<< ${(pj:\n:)tcandidates})"}) ;
 User: 0.00s    System: 0.00s  Total: 0.003s
 CPU:  94%      Mem:    8 MB
```

romkatv:
How about now? I've changed the code to use `sort` command if it exists and doesn't come from busybox. This is a pretty small code change and is a strict improvement over the current state. For everyone for whom fzf-tab used to work, it will continue to work and will have the same performance as before. For folks on busybox this code didn't work at all, and now it does (albeit slower than with GNU or BSD sort).

Aloxaf:
LGTM, Thanks.

romkatv:
Thanks!

PR #106, opened 2020-07-28
feat: add support of print-query

### LINKED ISSUE #102: [FR] Support the print-query fzf option
I would like to apply my input when no match was found. fzf provides the `print-query` option for this, but this plugin does not support it yet. 

**Is your feature request related to a problem? Please describe.**

The input disappears when no selection is applicable, and has to be re-typed again on the regular prompt. 
An example use-case is trying to auto-complete a path, and wanting to mkdir it if it doesn't exist yet. It would be useful to be able to just accept the non-matching input so mkdir can be appended.

When using the `--print-query` or the `--bind=ctrl-d:print-query` option for fzf, an error is thrown:

`_fzf_tab_complete:54: bad set of key/value pairs for associative array`

I assume this is because the printed query is not part of the `_fzf_tab_compcap` array.

Aloxaf:
Now we can use `alt-enter` to use the user input as the final result

Close #102

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Replace zsh's default completion selection menu with fzf!
         - [command](#command)
         - [extra-opts](#extra-opts)
         - [continuous-trigger](#continuous-trigger)
+        - [print-query](#print-query)
         - [ignore](#ignore)
         - [fake-compadd](#fake-compadd)
         - [insert-space](#insert-space)
@@ -156,13 +157,14 @@ Default value:
 FZF_TAB_COMMAND=(
     fzf
     --ansi   # Enable ANSI color support, necessary for showing groups
-    --expect='$continuous_trigger' # For continuous completion
+    --expect='$continuous_trigger,$print_query' # For continuous completion and print query
     '--color=hl:$(( $#headers == 0 ? 108 : 255 ))'
     --nth=2,3 --delimiter='\x00'  # Don't search prefix
     --layout=reverse --height='${FZF_TMUX_HEIGHT:=75%}'
     --tiebreak=begin -m --bind=tab:down,btab:up,change:top,ctrl-space:toggle --cycle
     '--query=$query'   # $query will be expanded to query string at runtime.
     '--header-lines=$#headers' # $#headers will be expanded to lines of headers at runtime
+    --print-query
 )
 zstyle ':fzf-tab:*' command $FZF_TAB_COMMAND
 ```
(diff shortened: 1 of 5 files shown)


NightMachinary:
Currently, this adds an additional keyboard shortcut to just print the query. While this is a good feature, it'd  better if it'd automatically print the query by the normal `enter` key when there is no selection possible in fzf:
image

Aloxaf:
@NightMachinary 

> it'd better if it'd automatically print the query by the normal enter key when there is no selection possible in fzf

Good idea.

Updated. How about now?

NightMachinary:
[**@Aloxaf**](https://github.com/Aloxaf) commented on [Jul 28, 2020, 5:11 PM GMT+4:30](https://github.com/Aloxaf/fzf-tab/pull/106#issuecomment-665016009 "2020-07-28T12:41:38Z - Replied by Github Reply Comments"):
> [@NightMachinary](https://github.com/NightMachinary)
> 
> > it'd better if it'd automatically print the query by the normal enter key when there is no selection possible in fzf
> 
> Good idea.
> 
> Updated. How about now?

Yes, it works perfectly (in my few tests :D). Thanks! This tool is a serious boost to the terminal UI. ❤️

NightMachinary:
@Aloxaf I have found a bug, I don't know if it's because of this new patch or not (I guess it is).

1. Start a completion by pressing tab:
image

2. Cancel it by pressing ESC.
3. What you had typed before pressing tab is now gone:
image
4. Type another character (say, `j`). Mysteriously, the space gets erased, too:
image

Aloxaf:
Oh yes. Thank you very much.

It has been fixed now.

NightMachinary:
[**@Aloxaf**](https://github.com/Aloxaf) commented on [Jul 29, 2020, 5:46 AM GMT+4:30](https://github.com/Aloxaf/fzf-tab/pull/106#issuecomment-665374731 "2020-07-29T01:16:07Z - Replied by Github Reply Comments"):
> Oh yes. Thank you very much.
> 
> It has been fixed now.

The bug is still present, but needs more tweaking to show itself. Try completing part of a path (using `/`), and then entering a nonexistent choice and pressing enter. It'll wipe everything. There was another similar bug, too, but I can't reproduce it reliably. It involved canceling the completion two times.

Aloxaf:
This is another bug: user input has no extra args which can tell compsys what should be added/removed from around the candidates.

I use the first candidate(generated by compsys)'s args as user input's, it should work in most cases. But sometimes it will also cause wrong result: `man \tabcdefg` will get `man 0 abcdefg`.

Issue #72, opened 2020-04-05
nvim -t infinite loop


MenkeTechnologies:
**Describe the bug**
nvim -t <tab> leads to infinite loop of
_tags:comptags:36: nesting level too deep
_tags:comptags:60: nesting level too deep
_tags:comptags:67: nesting level too deep

**To Reproduce**
1. Type 'nvim -t '
2. Press <kbd>Tab</kbd>

**Expected behavior**
tags will be in menucompletioon

**Screenshots**
<img width="279" alt="Screen Shot 2020-04-05 at 12 47 29 AM" src="https://user-images.githubusercontent.com/18195247/78467087-1294a880-76d7-11ea-98aa-1acf23c94c40.png">

**Environment (please complete the following information):**
 - OS: macoS
 - zsh version 5.8
 - fzf-tab version [e.g. https://github.com/Aloxaf/fzf-tab/commit/56adfa43b6cab964249c740302c8b57d129201f0]

**zshrc**
If applicable, add a minimal zshrc to help us analyze.

Aloxaf:
I can't reproduce this. Can you provide your zshrc?

Does `vim -t` have the same problem?

MenkeTechnologies:
good to know its on my end.  my zshrc is a little crazy.  i have a framework called zpwr and incorporated fzf-tab as a plugin.

https://raw.githubusercontent.com/MenkeTechnologies/zpwr/master/install/.zshrcv

Aloxaf:
That's really crazy and hard to debug.

Can you run the code below and then try to reproduce the bug and then provide the output of /tmp/fzf-tab-debug.txt?

```zsh
select-first() {
  read input
  echo $input
}

zstyle ':fzf-tab:*' command select-first

fzf-tab-debug() {
  local -i debug_fd=1
  exec {debug_fd}>&2 2>|/tmp/fzf-tab-debug.txt
  setopt xtrace
  local -i _fzf_tab_continue=1
  while (( _fzf_tab_continue )); do
    _fzf_tab_continue=0
    IN_FZF_TAB=1
    {
      zle .fzf-tab-orig-$_fzf_tab_orig_widget
    } always {
      IN_FZF_TAB=0
    }
    zle redisplay
  done
  unsetopt xtrace
}
zle -N fzf-tab-debug
bindkey "^I" fzf-tab-debug
```

NightMachinary:
I can't reproduce this with `nvim -t` either.

[...]

PR #122, opened 2020-09-23
fix _complete_tag calling _main_complete with args

MenkeTechnologies:
this is the root cause of https://github.com/Aloxaf/fzf-tab/issues/72.

ZSH lib file `_complete_tag`:

```sh
_main_complete - '' _wanted etags expl 'emacs tags' \
      compadd -a c_tags_array
elif [[ -f $c_path$c_tagsfile ]]; then
  # tags doesn't have as much in, but the tag is easy to find.
  # we can use awk here.
  c_tags_array=($(awk '{ print $1 }' $c_path$c_tagsfile))
  _main_complete - '' _wanted vtags expl 'vi tags' compadd -a c_tags_array
```

when fzf-tab hooks _main_complete, it must pass args to original function or else infinite recursion results when calling _main_complete without args.

ZSH lib file `_main_complete` uses positional arguments:
```
if (( $# )); then
  if [[ "$1" = - ]]; then
    if [[ $# -lt 3 ]]; then
      _completers=()
    else
      _completers=( "$2" )
      call=yes
    fi
  else
    _completers=( "$@" )
  fi
else
  zstyle -a ":completion:${curcontext}:" completer _completers ||
      _completers=( _complete _ignored )
fi
```

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -384,7 +384,7 @@ _fzf_tab_complete() {
     local -Ua _fzf_tab_groups
     local choice choices _fzf_tab_curcontext continuous_trigger ignore bs=$'\2' nul=$'\0'
 
-    _fzf_tab__main_complete  # must run with user options; don't move `emulate -L zsh` above this line
+    _fzf_tab__main_complete "$@" # must run with user options; don't move `emulate -L zsh` above this line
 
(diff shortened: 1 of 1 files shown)


Aloxaf:
Thanks a lot!

PR #123, opened 2020-09-27
list-colors ${(s.:.)LS_COLORS}

2moe:
Before:

Now:


Aloxaf:
Thanks for your contribution!

But I think this should be configured by the user themselves, as `${(s.:.)LS_COLORS}` is not the only possible value of `list-colors`.

PR #125, opened 2020-10-07
Check tab groups exist before hiding

rcarriga:
When `show-group` is set to brief, an error is raised by commands that don't provide a group. This happens in particular for commands using bash completion scripts like `aws`.
```sh
_fzf_tab_get_candidates:105: _fzf_tab_groups: assignment to invalid subscript range
```

This just adds a simple check for if the list is empty before accessing it


Aloxaf:
Thanks!

PR #126, opened 2020-10-17
feat: more powerful preview

Aloxaf:
This will close #79.

Example:

```zsh
zstyle ':fzf-tab:complete:*' extra-opts --preview=$fzf_tab_preview_init'
echo word: $word
echo description: $desc
echo group: $group
if (( $+ctxt[isfile] )); then
  echo path: $realpath
fi
'
```

PR #127, opened 2020-10-19
wip: improve completion description

adelin-b:
This pr aim to have a way to customize the description field
image

There is a lot of problem that I do not manage to resolve due to zsh-ing code, maybe you can give an helping hand, 
I wrote all the TODO inside the code for context
```zsh
# TODO once the dscr changed with the icon the $in variable in fzf preview has the icon in front,
#
# > echo $in 
#  copy_from_zsh_src.zsh
#
# how to get $word and $dscr in fzf preview
typeset -g fzf_tab_preview_init="
# TODO dscr
local dscr=\${\${\"\$(<{f})\"%\$'\0'*}#*\$'\0'*\$'\0'}
# TODO word
local word=\${\${\"\$(<{f})\"%\$'\0'*}#*\$'\0'*\$'\0'}
```

```zsh
# TODO how to get groups?
# GROUPS seem unaccessible
groups=
# TODO how to get group id ?
# I dont think it is safe if there is numbers before group
local gid="${__tmp_value//[!0-9]/}"
# TODO how to get group name ?
local group=groups[gid]

# TODO do this the same way of fzf_tab_preview_init
# # Here will be the custom script injection for icons
# if [ "$fzf_tab_completion_description_init" ]
#   dscr=$(eval 'local $dscr'$fzf_tab_completion_description_init)

# TODO remove this and use custom fzf_tab_completion_description_init above
dscr=$(echo "$dscr" | devicon-lookup)

```

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -361,7 +361,7 @@ _fzf_tab_get_candidates() {
         # add character and color to describe the type of the files
         dsuf='' dpre=''
         if (( $+v[isfile] )); then
-            filepath=${v[IPREFIX]}${v[hpre]}${k#*$'\b'}
+            filepath=${v[IPREFIX]}${v[hpre]}$v[word]
             filepath=${(Q)${(e)~filepath}}
             if (( $#list_colors && $+builtins[fzf-tab-colorize] )); then
               fzf-tab-colorize $filepath 2>/dev/null
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -91,6 +91,12 @@ _fzf_tab_compadd() {
     _opts+=("${(@kv)apre}" "${(@kv)hpre}" $isfile)
     __tmp_value+=$'\0args\0'${(pj:\1:)_opts}
 
+
+    group=$expl
+    groupid=$_fzf_tab_groups[(ie)$expl]
+
+    fzf_tab_compadd_hook
+
     # dscr - the string to show to users
     # example dscr:
     # Folder/
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -14,35 +14,18 @@ FZF_TAB_HOME=${0:h}
 
 source ${0:h}/lib/zsh-ls-colors/ls-colors.zsh fzf-tab-lscolors
 
-# This is an attenmpt at escaping fully fzf_tab_preview_init
-# typeset -g fzf_tab_preview_init='
-# # TODO once the dscr changed with the icon the $in variable in fzf preview has the icon in front,
-# # we need to provide hit, description, group to fzf_tab_preview_init
-# # trim input
-# local in=${${"$(<{f})\\"%$\\"\\0\\"*}#*$\\"\\0\\"*$\\"\\0\\"}
-# # get ctxt for current completion
-# local -A ctxt=("${(@ps:2:)CTXT}")
-# # get group
-# local -a groups=("${(@ps:2:)GROUPS}")
-# local -i gid=${"$(<{f})"%%$\\"\\0\\"*}
-# local group=$groups[gid]
-# # real path
-# local realpath=${ctxt[IPREFIX]}${ctxt[hpre]}$word
-# realpath=${(Qe)~realpath}
-# '
-
-typeset -g fzf_tab_preview_init="
+typeset -g fzf_tab_preview_init=$'
 # trim input
-local in=\${\${\"\$(<{f})\"%\$'\0'*}#*\$'\0'*\$'\0'}
+local in=${${"$(<{f})"%$\'\\0\'*}#*$\'\\0\'*$\'\\0\'}
 # get ctxt for current completion
-local -A ctxt=(\"\${(@ps:\2:)CTXT}\")
+local -A ctxt=("${(@ps:\\2:)CTXT}")
 # get group
-local -a groups=(\"\${(@ps:\2:)GROUPS}\")
-local -i gid=\${\"\$(<{f})\"%%\$'\0'*}
-local group=\$groups[gid]
+local -a groups=("${(@ps:\\2:)GROUPS}")
+local -i gid=${"$(<{f})"%%$\'\\0\'*}
+local group=$groups[gid]
 # real path
-local realpath=\${(Qe)~\${:-\${ctxt[IPREFIX]}\${ctxt[hpre]}}}\${(Q)in}
-"
+local realpath=${(Qe)~${:-${ctxt[IPREFIX]}${ctxt[hpre]}}}${(Q)in}
+'
 
 _fzf_tab_debug() {
   echo -E $'\n'${(qqqq)1}$'\n'
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -80,6 +80,8 @@ _fzf_tab_compadd() {
     _opts+=("${(@kv)apre}" "${(@kv)hpre}" $isfile)
     __tmp_value+=$'\0args\0'${(pj:\1:)_opts}
 
+    # Hook defined by user to alter the description of the completion
+    fzf_tab_compadd_hook
 
     # dscr - the string to show to users
     # example dscr:
@@ -98,9 +100,6 @@ _fzf_tab_compadd() {
         elif [[ -n $word ]]; then
             dscr=$word
         fi
-
-        # Hook defined by user to alter the description of the completion
-        fzf_tab_compadd_hook
 
         _fzf_tab_compcap+=$dscr$'\2'$__tmp_value${word:+$'\0word\0'$word}
     done
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -14,18 +14,24 @@ FZF_TAB_HOME=${0:h}
 
 source ${0:h}/lib/zsh-ls-colors/ls-colors.zsh fzf-tab-lscolors
 
-# TODO once the dscr changed with the icon the $in variable in fzf preview has the icon in front,
-#
-# > echo $in 
-#  copy_from_zsh_src.zsh
-#
-# how to get $word and $dscr in fzf preview
-typeset -g fzf_tab_preview_init="
-# TODO dscr
-local dscr=\${\${\"\$(<{f})\"%\$'\0'*}#*\$'\0'*\$'\0'}
-# TODO word
-local word=\${\${\"\$(<{f})\"%\$'\0'*}#*\$'\0'*\$'\0'}
+# This is an attenmpt at escaping fully fzf_tab_preview_init
+# typeset -g fzf_tab_preview_init='
+# # TODO once the dscr changed with the icon the $in variable in fzf preview has the icon in front,
+# # we need to provide hit, description, group to fzf_tab_preview_init
+# # trim input
+# local in=${${"$(<{f})\\"%$\\"\\0\\"*}#*$\\"\\0\\"*$\\"\\0\\"}
+# # get ctxt for current completion
+# local -A ctxt=("${(@ps:2:)CTXT}")
+# # get group
+# local -a groups=("${(@ps:2:)GROUPS}")
+# local -i gid=${"$(<{f})"%%$\\"\\0\\"*}
+# local group=$groups[gid]
+# # real path
+# local realpath=${ctxt[IPREFIX]}${ctxt[hpre]}$word
+# realpath=${(Qe)~realpath}
+# '
 
+typeset -g fzf_tab_preview_init="
 # trim input
 local in=\${\${\"\$(<{f})\"%\$'\0'*}#*\$'\0'*\$'\0'}
 # get ctxt for current completion
@@ -92,17 +98,14 @@ _fzf_tab_compadd() {
     __tmp_value+=$'\0args\0'${(pj:\1:)_opts}
 
 
-    group=$expl
-    groupid=$_fzf_tab_groups[(ie)$expl]
-
-    fzf_tab_compadd_hook
-
     # dscr - the string to show to users
     # example dscr:
     # Folder/
+    # 653fab6  -- [HEAD]    feat: add compadd hook (4 hours ago)
     # word - the string to be inserted
     # example word:
     # Documents/Folder/
+    # 653fab6
     local dscr word i
     
     for i in {1..$#__hits}; do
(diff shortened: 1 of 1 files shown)


Aloxaf:
That's a cool idea. But it can also be implemented without modifing the code of fzf-tab. (After I fix a colorizing bug)

图片

```zsh
# A dirty trick, I am not sure whether it will cause bug somewhere
_files() {
  # can also use `exa --icons` here
  local -a descriptions=(${(f)"$(command ls | devicon-lookup)"})
  local -a files=(${(f)"$(command ls)"})
  compadd -a -f -d descriptions -- files
}
```

adelin-b:
Interesting, I tryed it and I didn't though it was so simple to have icons on normal completion, in that snippet the hidden files are missing when completing.

How would that go for git command for example, I had in mind to show some icon/status in front of the completed file
image

This is what I got based on your snippet 
```
_git-add() {
  # can also use `exa --icons` here
  local -a descriptions=(${(f)"$(command exa --icons --git --long --color=always)"})
  local -a files=(${(f)"$(command ls)"})
  compadd -a -f -d descriptions -- files
}
```
image


But I wonder how to get that file status for all git subcommand completions that involve files.

With `$group` all I have to do was to match `*file` as for completion this way I think it need to be defined for each subcommand, and for complex completions the original completion function need to be reimplemented.
Instead of 
`completion -> fzf-tab`
Something like this seem easier to plug on existing completions
`completion -> fzf-tab-description-init -> fzf-tab`

Based on the preview pipeline that you changed in !126
`completion-> fzf-tab -> preview`
`completion-> fzf-tab -> fzf-tab-preview-init -> preview`

adelin-b:
Another example of workflow to add icons to the `command` completion
image
First I need to get where the completion come from use `ctrl-x h`
```zsh
❯ ls
tags in context :completion::complete:-command-::
    commands builtins functions aliases suffix-aliases reserved-words jobs parameters parameters  (_command_names _autocd) 
    commands                                                                                      (_path_commands _command_names _autocd) 
    jobs                                                                                          (_jobs _command_names _autocd) 
    parameters                                                                                    (_parameters _command_names _autocd) 
tags in context :completion::files:::
    globbed-files  (_files _fzf_tab__main_complete _fzf_tab_complete)
```

from this at some point I need to understand that I need to override the `_command_names`
```zsh
❯ which _command_names
_command_names () {
	local args defs ffilt
	zstyle -t ":completion:${curcontext}:commands" rehash && rehash
	zstyle -t ":completion:${curcontext}:functions" prefix-needed && [[ $PREFIX != [_.]* ]] && ffilt='[(I)[^_.]*]' 
	defs=('commands:external command:_path_commands') 
	[[ -n "$path[(r).]" || $PREFIX = */* ]] && defs+=('executables:executable file:_files -g \*\(-\*\)') 
	if [[ "$1" = -e ]]
	then
		shift
	elif (( ${#precommands:|builtin_precommands} ))
	then
		
	else
		[[ "$1" = - ]] && shift
		defs=("$defs[@]" 'builtins:builtin command:compadd -Qk builtins' "functions:shell function:compadd -k 'functions$ffilt'" 'aliases:alias:compadd -Qk aliases' 'suffix-aliases:suffix alias:_suffix_alias_files' 'reserved-words:reserved word:compadd -Qk reswords' 'jobs:: _jobs -t' 'parameters:: _parameters -g "^*(readonly|association)*" -qS= -r "\n\t\- =[+"' 'parameters:: _parameters -g "
*association*~*readonly*" -qS\[ -r "\n\t\- =[+"') 
	fi
	args=("$@") 
	local -a cmdpath
	if zstyle -a ":completion:${curcontext}" command-path cmdpath && [[ $#cmdpath -gt 0 ]]
	then
		local -a +h path
		local -A +h commands
		path=($cmdpath) 
	fi
	_alternative -O args "$defs[@]"
}
```
And I need to hack into this 300+ character long zsh blob line to provide the icon.

With my proposal it would just be someting like
```zsh
case $group in
    (function)
        icon=f
    (alias)
        icon=a
    (*)
        icon=_>
esac
description=$icon$description
```

Aloxaf:
Umm, that's reasonable.

But process these descriptions one by one by calling a function is inefficient. A better way may be add something like `fzf_tab_get_candidates_hook`.

Aloxaf:
I find a new way again. We can add something like `fzf_compadd_hook`.

In this case, words are stored in `$__hits`, old description in `$__dscr`, group name is `$expl`(They may need a better name).

What do you think?
```diff
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -63,6 +63,11 @@ _fzf_tab_compadd() {
     _opts+=("${(@kv)apre}" "${(@kv)hpre}" $isfile)
     __tmp_value+=$'\0args\0'${(pj:\1:)_opts}

+    # add icon
+    if [[ -n $isfile ]]; then
+      __dscr=(${(f)"$(devicon-lookup <<< ${(pj:\n:)__hits})"})
+    fi
+
     # dscr - the string to show to users
     # word - the string to be inserted
     local dscr word i
```

adelin-b:
I do not understand a single line with zsh magic in it, I will try this this evening. 

The aim is to do that by batch in that case if I understand, this is great. 

Here is a few use cases I wanted to implement by messing with the completion description I don't think it can be alway done by batch tho. 
- show docker status
- show systemd services status 
- show snippets that match from tldr (will add new hits) 
- show snippets that match history (will add hits) 
- show variable content when completing them

If it could be a very similar api as fzf_tab_preview_init so we can use `switch case` matching every group that would be perfect.

Aloxaf:
> The aim is to do that by batch in that case if I understand, this is great.

Yes, it is.

> 
> Here is a few use cases I wanted to implement by messing with the completion description I don't think it can be alway done by batch tho.

No. You can also process them one by one if you want.

For example, you can do something like this:

```zsh
# this is very inefficent, just a demo
fzf_tab_compadd_hook() {
  if [[ $_fzf_tab_curcontext == *systemctl-start* ]]; then
      __dscr=()
      for service in $__hits; do
	      if [[ $(systemctl status $service) == *running* ]]; then
		      __dscr+=("[O]$service")
  		  else
			  __dscr+=("[X]$service")
   		  fi
      done
  fi
}
```

图片


This way is also easy to implement, just check and run `fzf_tab_compadd_hook` if it exists is enough

adelin-b:
So I tryed it and it is nice for batch and adding custom completion on the go 

The part misssing to me now is how to make custom preview on `$group` instead of the completion context, because `$_fzf_tab_curcontext` dont allow to be more specific than this

adelin-b:
The current issue is that the description only is in the preview (with icon/ text and such), while we want also the origin $hit to apply preview on it. Im quite lost on this part

I had to try modify the fzf_tab_preview_init because it is not escaped but unsucessfully, I did put a comment with the attempt using single quote ' but it conflict with the other single quote 

@Aloxaf you can test with this in the .zshrc, it use the nerd icon mostly, I dont know how to make it universal for now.

```
fzf_tab_compadd_hook() {
  case $_fzf_tab_curcontext in
    (*)
      case $expl in
        (function) icon="ƒ" ;;
        (alias) icon="Ą" ;;
        (*branch*) icon="" ;;
        (*command*) icon="" ;;
        (*file*) icon=$(echo $hit | devicon-lookup) hit="" ;;
        (*target*) icon="" ;;
        (*option*)
          case $hit in
            (*rm*|*delete*|*remove*)
              icon=" " ;;
            (*help) icon="" ;; # icon=" " icon=" " icon=""
            (*) icon="" ;; # icon=" "
          esac
          ;;
        (*) icon=" " ;;
      esac
      dscr=("$icon $dscr")
      ;;
  esac
}

```

adelin-b:
I really think all the magic strings should be broked down into well named variables, as it is impossible to maintain easily
Then for performance purpose concatening them and commenting the explicit code that would show the breakdown of if

Aloxaf:
> The current issue is that the description only is in the preview (with icon/ text and such), while we want also the origin $hit to apply preview on it. Im quite lost on this part

See https://github.com/Aloxaf/fzf-tab/pull/127#issuecomment-714178829

fzf-tab.zsh
```diff
@@ -80,19 +97,33 @@ _fzf_tab_compadd() {
...
     for i in {1..$#__hits}; do
         word=$__hits[i] dscr=$__dscr[i]
         if [[ -n $dscr ]]; then
             dscr=${dscr//$'\n'}
         elif [[ -n $word ]]; then
             dscr=$word
         fi
+
+        # Hook defined by user to alter the description of the completion
+        fzf_tab_compadd_hook
```
Aloxaf, line 103: Please don't call `fzf_tab_compadd_hook` in the loop, this is very low inefficient.

Just put it before this loop. And check whether it exists before calling it (You can use `$+functions[fzf_tab_compadd_hook]` here).

Aloxaf:
> I really think all the magic strings should be broked down into well named variables, as it is impossible to maintain easily

Yes, you can bind them to a well-named variable temporary.

Like this:
```zsh
() {
  local group=$expl
  fzf_tab_compadd_hook
}
# group is unset here.
```

fzf-tab.zsh
adelin-b: The issue when doing this was that when out of the loop I didnt managed to get the original hits + description pair

for example
```
fzf_tab_compadd_hook() {
  if [[ $_fzf_tab_curcontext == *systemctl-* ]]; then
      __dscr=()
      for service in $__hits; do
	      if [[ $(systemctl status $service) == *inactive* ]]; then
		      __dscr+=("[X]$service")
  		  else
			  __dscr+=("[O]$service")
   		  fi
      done
  fi
}
```
For a systemctl status it work but when the description is bigger (for a commit) this snippet loose the original description.

adelin-b: When the hook is above the loop here is what I was using, but loosing the completion. And I didnt found a way to make it batch well at the moment and ended up with another loop
```zsh
fzf_tab_compadd_hook() {

    # if [[ -n $isfile ]]; then
    #   __dscr=(${(f)"$(devicon-lookup <<< ${(pj:\n:)__hits})"})
    # fi
    case $_fzf_tab_curcontext in
      (*)
        __dscr=()
        for hit in $__hits; do
          case $group in
            (function) icon="ƒ" ;;
            (alias) icon="Ą" ;;
            (*branch*) icon="" ;;
            (*command*) icon="" ;;
            (*file*) icon=$(echo $hit | devicon-lookup) hit="" ;;
            (*target*) icon="" ;;
            (*option*)
              case $hit in
                (*rm*|*delete**remove*)
                  icon=" " ;;
                (*help) icon="" ;; # icon=" " icon=" " icon=""
                (*) icon="" ;; # icon=" "
              esac
              ;;
            (*) icon=" " ;;
          esac

          __dscr+=("$icon $hit")
        done
        ;;
    esac
  }

```

Aloxaf: > For a systemctl status it work but when the description is bigger (for a commit) this snippet loose the original description.

My code is just a demo. In your case, you should edit the description in place.

```
for i in {1..$#__hits}; do
    __dscr[i]=$SOME_ICON$__dscr[i]
done
```

adelin-b: I dont even get ho to assign a value to __dscr[i] without fucking it up
```zsh
fzf_tab_compadd_hook() {
  SOME_ICON="ƒ"
  local i
  for i in {1..$#__hits}; do
    # __dscr[i]=$SOME_ICON$__dscr[i]
    # __dscr[i]=${__dscr[i]}
    # $__dscr[i]="hello"
    # ${__dscr[i]}="hello"
    # __dscr[i]="hello"
    # __dscr[i]="hello "
    # __dscr[i]="hello \n"
  done
}
```

adelin-b:
> > I really think all the magic strings should be broked down into well named variables, as it is impossible to maintain easily
> 
> Yes, you can bind them to a well-named variable temporary.
> 
> Like this:
> 
> ```shell
> () {
>   local group=$expl
>   fzf_tab_compadd_hook
> }
> # group is unset here.
> ```

By magic lines I was more thinking about stuff like `local in=${${"$(<{f})"%$\'\\0\'*}#*$\'\\0\'*$\'\\0\'}`

fzf-tab.zsh
Aloxaf: Sometimes `$__dscr` is empty (eg. file name has no description). In this case, you should use `$__hits` instead.

```zsh
if (( $#__dscr )); then
    for i in {1..$#__hits}; do
        __dscr[i]="[O] "$__dscr[i]
    done
else
    for i in {1..$#__hits}; do
        __dscr[i]="[O] "$__hits[i]
    done
fi
```

Aloxaf:
> By magic lines I was more thinking about stuff like `local in=${${"$(<{f})"%$\'\\0\'*}#*$\'\\0\'*$\'\\0\'}`

It turn the description from internal representation in fzf-tab into a normal string. It's equivalent to `local in=$(cut -d $'\0' -f2 <<< {f})` I think.

BTW, users dont't need to and should not manipulate with it as it may change in the future.

adelin-b:
Okay I think im done with the logic, the only problem left atm is that the preview get the `description` and not the `word` so it crash when there is icons added or special characters
```
fzf_tab_compadd_hook() {
  case $_fzf_tab_curcontext in
    # All context
    (*)
    local already_enriched=0
    # Batch lookup icon assignements
    if [[ -n $isfile ]]; then
      __dscr=(${(f)"$(devicon-lookup <<< ${(pj:\n:)__hits})"})
      already_enriched=1
    fi

        # Line by line icon assignements
        if [[ already_enriched == 1 ]]; then
          for i in {1..$#__hits}; do
            hit=$__hits[i]
            word=$__hits[i] dscr=$__dscr[i]
            if [[ -n $dscr ]]; then
              dscr=${dscr//$'\n'}
            elif [[ -n $word ]]; then
              dscr=$word
            fi

            case "$expl" in
              (function) icon="ƒ" ;;
              (alias) icon="Ą" ;;
              (*branch*) icon="" ;;
              (*command*) icon="" ;;
              # (*file*) icon=$(echo $hit | devicon-lookup);;
            (*target*) icon="" ;;
            (*option*)
              case $hit in
                (*rm*|*delete**remove*)
                  icon=" " ;;
                (*help) icon="" ;; # icon=" " icon=" " icon=""
                (*) icon="" ;; # icon=" "
              esac
              ;;
            (*) icon="" ;;
          esac
          __dscr[i]=$icon$dscr
        done
        fi
        ;;
    esac
  }
```

Aloxaf:
> Okay I think im done with the logic, the only problem left atm is that the preview get the `description` and not the `word` so it crash when there is icons added or special characters

Now you can: [`4047009` (#126)](https://github.com/Aloxaf/fzf-tab/pull/126/commits/40470092573a39bcc065c8415d060ad01f1b6e31)

Aloxaf:
Well, this PR is closed automatically after I merged #126.

You can make a new PR to the master branch.

PR #130, opened 2020-10-21
feat: add support for tmux popup

### LINKED ISSUE #128: [FR] integration with tmux popup
https://github.com/tmux/tmux/issues/1842

That would be useful if the current pane is too small to show the full candidate list.

Aloxaf:
[asciicast](https://asciinema.org/a/yiyBrosZ3OmgkNRJ3NLxObUe3)

I love this. It won't push my prompt to the top of pane when the completion list is too long.

Close #128.

How to use (**need tmux 3.2!!**):

```zsh
FZF_TAB_COMMAND[1]=fzf-tmux-popup
zstyle ':fzf-tab:*' command $FZF_TAB_COMMAND
```

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -148,6 +148,7 @@ _fzf_tab_get() {
     _fzf_tab_add_default group-colors $FZF_TAB_GROUP_COLORS
     _fzf_tab_add_default ignore false
     _fzf_tab_add_default print-query alt-enter
+    _fzf_tab_add_default popup-pad 0 0
 
     if zstyle -m ':completion:*:descriptions' format '*'; then
         _fzf_tab_add_default prefix '·'
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -12,6 +12,7 @@ zmodload -F zsh/stat b:zstat
 0="${${(M)0:#/*}:-$PWD/$0}"
 FZF_TAB_HOME=${0:h}
 
+autoload -Uz $FZF_TAB_HOME/fzf-tmux-popup
 source ${0:h}/lib/zsh-ls-colors/ls-colors.zsh fzf-tab-lscolors
 
 # thanks Valodim/zsh-capture-completion
--- a/fzf-tmux-popup
+++ b/fzf-tmux-popup
@@ -1,4 +1,5 @@
-#!/usr/bin/env zsh
+#!/hint/zsh
+emulate -L zsh -o extended_glob
 
 local -a fzf_opts=($@)
 fzf_opts=(${${fzf_opts/--height*}/--layout*})
@@ -14,7 +15,16 @@ cat > /tmp/fzf-tab/list-$$
 
 # get the size of content, note we should remove all ANSI color code
 local comp_lines=${#${(f)"$(</tmp/fzf-tab/list-$$)"}}
-local comp_length=$(sed 's/\x1b\[[0-9;]*m//g' /tmp/fzf-tab/list-$$ | awk 'length > max_length { max_length = length; } END { print max_length }')
+# FIXME: can't get the correct width of CJK characters.
+if (( comp_lines <= 500 )); then
+  local comp_length=0 length
+  for line in ${(f)"$(</tmp/fzf-tab/list-$$)"}; do
+    length=${#${(S)line//$'\x1b['[0-9]#*m}}
+    (( length >= comp_length )) && comp_length=$length
+  done
+else
+  local comp_length=$(sed 's/\x1b\[[0-9;]*m//g' /tmp/fzf-tab/list-$$ | awk 'length > max_length { max_length = length; } END { print max_length }')
+fi
 
 # calculate the popup height and y position
 if (( cursor_y * 2 > window_height )); then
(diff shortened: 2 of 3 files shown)


solarizedalias:
This is nice! I like it.
Do you think it's possible to decide the width taking into account the preview window too?
It may be difficult to know exactly how many characters the width of preview window would take.

Perhaps it's better to configure it with smaller height and bigger width but that's my preference.
So it will be more awesome if it's configurable by zstyle for specific commands, context.

Aloxaf:
> Do you think it's possible to decide the width taking into account the preview window too?
> It may be difficult to know exactly how many characters the width of preview window would take.
> 
> Perhaps it's better to configure it with smaller height and bigger width but that's my preference.

I am also looking for a way to support the preview window.

Two possible solution:
1. Move the preview window to the bottom.
2. Add a configurable variable `FZF_TAB_TMUX_POPUP_PADDING`.

> So it will be more awesome if it's configurable by zstyle for specific commands, context.

Yes it is. `command` tag is configurable. For example, you can enable it just for command options:
```zsh
zstyle ':fzf-tab:*:options' command $FZF_TAB_COMMAND_POPUP
```

solarizedalias:
A few things I've noticed.

* Inside the popup, PWD is current directory of the tmux session rather than the path of the pane which invoked the `fzf-tmux-popup`. This breaks some extra config like `--preview=$extract' ls ${realpath}'` from the README.md examples. This would be easily fixed with something like this.

	```zsh
	local pane_cwd
	pane_cwd=$(tmux display-message -p "#{pane_current_path}")
	 ...
	# and then
	echo -E "cd ${pane_cwd} && fzf ${(qq)fzf_opts[@]} < /tmp/fzf-tab/list-$$ > /tmp/fzf-tab/result-$$" > /tmp/fzf-tab/fzf-$$
	```

* It looks like the descriptions are truncated to the pane's width so you will still lose the exceeded lengths even if you set a larger widths for popup. It does not affect the long candidates or the preview window so I'm guessing this is done by zsh's compsys feature so that the long descriptions are trimmed according to `$COLUMNS` of the pane width. (Note that I set popup width lager than the pane widths in the recording.)

	[asciicast](https://asciinema.org/a/u2H6ropdqTdr0zAFKLESKpRu5)

And here's what I modified mentioned all the above.

```patch
diff --git a/fzf-tmux-popup b/fzf-tmux-popup
index 69f6af5..43c26e1 100755
--- a/fzf-tmux-popup
+++ b/fzf-tmux-popup
@@ -6,6 +6,9 @@ fzf_opts=(${${fzf_opts/--height*}/--layout*})
 
 # get position of cursor and size of window
 local -a tmp=($(tmux display-message -p "#{pane_top} #{cursor_y} #{pane_left} #{cursor_x} #{window_height} #{window_width}"))
+local pane_cwd
+pane_cwd=$(tmux display-message -p "#{pane_current_path}")
+
 local cursor_y=$((tmp[1] + tmp[2])) cursor_x=$((tmp[3] + tmp[4])) window_height=$tmp[5] window_width=$tmp[6]
 
 # create fifo and write to completion lists to file
@@ -18,8 +21,8 @@ local comp_lines=${#${(f)"$(</tmp/fzf-tab/list-$$)"}}
 # FIXME: can't get the correct width of CJK characters.
 if (( comp_lines <= 500 )); then
   local comp_length=0 length
-  for line in ${(f)"$(</tmp/fzf-tab/list-$$)"}; do
-    length=${#${(S)line//$'\x1b['[0-9]#*m}}
+  for line in ${(@)${(f)"$(</tmp/fzf-tab/list-$$)"}//(#u)[^[:print:]]/}; do
+    length=${#${line}}
     (( length >= comp_length )) && comp_length=$length
   done
 else
@@ -30,19 +33,22 @@ fi
 if (( cursor_y * 2 > window_height )); then
   # show above the cursor
   local popup_height=$(( comp_lines >= cursor_y ? cursor_y : comp_lines + 4 ))
+  popup_height=$(( popup_height >= window_height / 2 ? (window_height / 2) : (popup_height) ))
   local popup_y=$cursor_y
 else
   # show below the cursor
   local popup_height=$(( comp_lines >= (window_height - cursor_y) ? window_height - cursor_y : comp_lines + 4 ))
-  local popup_y=$(( cursor_y + popup_height + 1 ))
+  popup_height=$(( popup_height >= window_height / 2 ? (window_height / 2) : (popup_height) ))
+  local popup_y=$(( cursor_y + popup_height + 2 ))
   fzf_opts+=(--layout=reverse)
 fi
 
 # calculate the popup width and x position
 local popup_width=$(( comp_length + 5 > window_width ? window_width : comp_length + 5 ))
+popup_width=$(( popup_width * 2 >= window_width ? popup_width : window_width / 2  ))
 local popup_x=$(( cursor_x + popup_width > window_width ? window_width - popup_width : cursor_x ))
 
-echo -E "fzf ${(qq)fzf_opts[@]} < /tmp/fzf-tab/list-$$ > /tmp/fzf-tab/result-$$" > /tmp/fzf-tab/fzf-$$
+echo -E "cd ${pane_cwd} && fzf ${(qq)fzf_opts[@]} < /tmp/fzf-tab/list-$$ > /tmp/fzf-tab/result-$$" > /tmp/fzf-tab/fzf-$$
 {
   tmux popup -x $popup_x -y $popup_y \
        -w $popup_width -h $popup_height \
```

solarizedalias:
Oh looks like it's merged while I'm editing. Should I move it to issues?

Aloxaf:
@solarizedalias Oh sorry I didn't notice your reply.

> Inside the popup, PWD is current directory of...

This bug is fixed in https://github.com/Aloxaf/fzf-tab/commit/966e7b47cd937014e8a0ec7de5273d3c60ffbeeb. I disabled all the preview while developing this branch so I didn't notice this bug but after merging it into master I soon notice it.

---

> It looks like the descriptions are truncated to the pane's width so you will still lose the exceeded lengths 

This is an known issue. But I still haven't find a way to solve it. I think I should move it into my TODO lists now.

---

> Oh looks like it's merged while I'm editing. Should I move it to issues?

Feel free to move it to issue.

PR #133, opened 2020-10-24
fix: realpath using word instead of description

adelin-b:
Following #127

This add an hook to be able to enrich the description of completion
image
image
Git status
image

## What is left to do
- [ ] How to set custom color ? it seem that if set in `$__dscr` they get escaped no matter what

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,74 @@ If it is a number, then fzf-tab won't be activated if the number of candidates i
 
 Default value: `zstyle ':fzf-tab:*' ignore false`
 
+### compadd-hook
+
+You can hook on completion value and improve it using `fzf_tab_compadd_hook` 
+
+For example to add a lot of icons on many completions (Nerd icons patched fonts)
+
+```zsh
+fzf_tab_compadd_hook() {
+  # $expl - explicit group name
+  # $__hits - list of orginal completions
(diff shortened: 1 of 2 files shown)


Aloxaf:
1. This is difficult... You can reset the color, but you have no way to restore it.

2. There is an array `$words` which contains the words on the command line. But it is editable so some completion function may remove some of its content...

adelin-b:
> 1. This is difficult... You can reset the color, but you have no way to restore it.

How can I do that ? it seems that the colors are getting escaped when I try to insert some inside the completion

Aloxaf:
> How can I do that ? it seems that the colors are getting escaped when I try to insert some inside the completion

You need to use $'\033[0m'

adelin-b:
Hello, I saw that this wasnt merged and Im still using the modified branch, I managed to do most of what I aimed for, beside color that did not work. However I found out you rewrote it all in C. 

Here is what I have at the moment on https://github.com/adelin-b/fzf-tab/tree/improve_completion_description
Animation

It is on the old code so It might get me a while to move it to C. Maybe you already implemented the way to do it exposing different variable name?

Aloxaf:
Oh sorry I haven't implement it in C. I will take a look at it in these days.

PR #145, opened 2020-10-30
Use command rm when possible

NICHOLAS85:
Many users have rm aliased to a different command or added flags such as `-I` or `-v`. Since these rm evocations aren't supposed to be verbose/visible, I'd find it best to force the use of the silent regular rm.


Aloxaf:
Thanks for your PR.

In fact, alias won't affect these functions since they are loaded with `autoload -U`， which will suppress alias expansion.

But I didn't consider the situation where the user defined a function called `rm`.

PR #148, opened 2020-11-02
feat: add support of accept-line

NICHOLAS85:
Allows a user to accept and run a suggestion in one keystroke.
https://asciinema.org/a/dMcynTzjKZRC0nZbvSo4gmIla

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -95,7 +95,7 @@
 -ftb-complete() {
   local -a _ftb_compcap
   local -Ua _ftb_groups
-  local choice choices _ftb_curcontext continuous_trigger print_query bs=$'\2' nul=$'\0'
+  local choice choices _ftb_curcontext continuous_trigger print_query accept_line bs=$'\2' nul=$'\0'
 
   # must run with user options; don't move `emulate -L zsh` above this line
   (( $+builtins[fzf-tab-compcap-generate] )) && fzf-tab-compcap-generate -i
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -116,6 +116,7 @@
       -ftb-generate-header     # sets `_ftb_headers`
       -ftb-zstyle -s continuous-trigger continuous_trigger || continuous_trigger=/
       -ftb-zstyle -s print-query print_query || print_query=alt-enter
+      -ftb-zstyle -s accept-line accept_line || accept_line=enter
 
       # NOTE: Using pipe here causes an error "failed to read /dev/tty"
       # when _ftb_complist is long
@@ -145,6 +146,10 @@
   if [[ $choices[1] && $choices[1] == $continuous_trigger ]]; then
     typeset -gi _ftb_continue=1
   fi
+
+  if [[ $choices[1] && $choices[1] == $accept_line ]]; then
+    typeset -gi _ftb_accept=1
+  fi
   choices[1]=()
 
   for choice in "$choices[@]"; do
@@ -170,6 +175,7 @@ fzf-tab-complete() {
   local -i _ftb_continue=1
   while (( _ftb_continue )); do
     _ftb_continue=0
+    _ftb_accept=0
     local IN_FZF_TAB=1
     {
       zle .fzf-tab-orig-$_ftb_orig_widget
@@ -181,7 +187,7 @@ fzf-tab-complete() {
       zle .reset-prompt
       zle -R
     else
-      zle redisplay
+      (( _ftb_accept )) && zle accept-line || zle redisplay
     fi
   done
 }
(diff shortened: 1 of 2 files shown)


Aloxaf:
Thanks for your PR. But I think this is not very useful.

Because you can't bind it to <kbd>enter</kbd>, which is already binded to `accept` event. I believe most people use it just for accept a completion.

And if we bind it to something like <kbd>ctrl+enter</kbd>, users will still need two keystroke.

NICHOLAS85:
I can see how it wouldn't be useful if someone bound it to something other than a single keystroke. Instead of creating a configurable keybind, like in this pr, would you consider being able to alternate the `accept` behavior with a setting? If you're curious about how often people would use this behavior, here's a pr and discussion https://github.com/junegunn/fzf/pull/1492 which references a similar request but for a different widget.

NICHOLAS85:
I've been testing creating a boolean option which changes the default accept behavior to accept or accept and run. However I found that by doing so, when the option is set, you lose the ability to accept without running. This draws me back to this pr which preserves the original behavior of accept and also allows a keybind to be set for accept and run. In my case I have the following set:
```zsh
zstyle ':fzf-tab:*' fzf-bindings 'space:accept'
zstyle ':fzf-tab:*' accept-line enter
```
Which works very well for me and allows me to choose between accept or accept and run

up-to-you:
i applied all changes and it works like a charm, my setup:

```
zstyle ':fzf-tab:*' continuous-trigger 'space'
zstyle ':fzf-tab:*' fzf-bindings 'tab:accept'
zstyle ':fzf-tab:*' accept-line enter
zstyle ':completion:complete:*:options' sort false
```

@Aloxaf hope it can be merged to upstream branch

tjx666:
@Aloxaf @NICHOLAS85 @up-to-you I want to set shift + enter to accept and excute, how can can configure it?

For [zsh-autosuggestions](https://github.com/zsh-users/zsh-autosuggestions), I just need to config:

```
bindkey '^[SE' autosuggest-execute
```

NICHOLAS85:
You will use the following substituting `$keybind`:

```zsh
zstyle ':fzf-tab:*' accept-line $keybind
```

Another thing to note, most terminals I've used don't differentiate between enter, ctrl+enter, and shift+enter by default. You may need to look into setting that up properly before you can use it as a bind: https://unix.stackexchange.com/questions/536352/ctrl-enter-shift-enter-and-enter-are-interpreted-as-the-same-key

tjx666:
@NICHOLAS85

I using the VSCode to send escaped code to shell, so the zsh-autosuggestion works.

image

tjx666:
I find following setting doesn't work, even the fzf window doesn't show

```sh
zstyle ':fzf-tab:*' accept-line '^[SE'
```


https://user-images.githubusercontent.com/41773861/196004611-8c5b0a2d-e398-4362-98c0-9eaccf5f4ef2.mov

NICHOLAS85:
I haven't messed with fzf in awhile but I believe it's due to there being hard coded supported keys: https://www.mankier.com/1/fzf#Key/Event_Bindings-Available_Keys:_(Synonyms)

You could try remapping the sent code to an unused binding in that list

tjx666:
@NICHOLAS85 
I think the problem is how to use character sequence as zstyle keybinding.
Seems `'^[SE'` is not valid keybinding, but using `space` is work.

Aloxaf:
@tjx666
The answer in  https://github.com/Aloxaf/fzf-tab/pull/148#issuecomment-1279843328 is clear enough.

Example: 

```json
[
    {
        "key": "shift+enter",
        "command": "workbench.action.terminal.sendSequence",
        "when": "terminalFocus && terminalShellIntegrationEnabled && !accessibilityModeEnabled",
        "args": {
            "text": "\u001b[1;3C"
        }
    },
]
```
Above lines remap `shift+enter` to `alt+right`.
Then you can use `zstyle ':fzf-tab:*' accept-line alt+right` to use `shift+enter` to accept and excute.

tjx666:
@Aloxaf 

设置 `zstyle ':fzf-tab:*' accept-line alt+right` 和上面视频里一样，按 tab 后直接换行了，fzf 的搜索框都没法打开

Aloxaf:
> @Aloxaf
> 
> 设置 `zstyle ':fzf-tab:*' accept-line alt+right` 和上面视频里一样，按 tab 后直接换行了，fzf 的搜索框都没法打开


Oh sorry, it's `alt-right`, not `alt+right`. And make sure you have vscode configured correctly.

tjx666:
For anyone find the solution to using `shift + enter` to accept and execute.

In vscode, you first need to config shortcut:

```json
{
        "key": "shift+enter",
        "command": "workbench.action.terminal.sendSequence",
        "when": "terminalFocus",
        "args": { "text": "\u001b[1;3B" }
}
```

Then set `zstyle ':fzf-tab:*' accept-line alt-down` to you .zshrc.

If you want your iterm2 work, follow the screenshot:

image

PR #153, opened 2020-11-10
fix: FZF_TAB_HOME when a symlink is sourced or the path contains a space

druckdev:
`FZF_TAB_HOME` will contain a wrong directory when instead of `fzf-tab.zsh` or `fzf-tab.plugin.zsh`, a symlink pointing to one of the two is sourced and the symlink lays in a different directory. In this case it will not contain the repository root but the directory in which the symlink is.
Errors also occur when the path of the repository contains spaces.

This fixes both of these issues by passing the path through `realpath` and quoting `$0`.


Aloxaf:
LGTM. Thanks a lot.

Aloxaf:
> Errors also occur when the path of the repository contains spaces.

This shouldn't happen. Have you set `SH_WORD_SPLIT`?

图片

PR #121, opened 2020-09-08
[Fix #112] Set DL_EXT=bundle if $OSTYPE is darwin*

solarizedalias:
This PR adds `DL_EXT=bundle` to the arguments of `./configure` when the OS is macOS (detected by `[[ ${OSTYPE} == darwin* ]]`), changing the extension of the compiled module from `.so` to `.bundle`, which `zmodload` would expect in macOS. This fixes #112

PR #159, opened 2020-11-17
fix: Check also .bundle as a module extension

solarizedalias:
This PR adds checking `.bundle` as a module extension (for macOS) in the initialization code.
I forgot to add this check in the previous PR https://github.com/Aloxaf/fzf-tab/pull/121#issue-481954020 about module build in macOS.


Aloxaf:
LGTM, Thanks

PR #183, opened 2021-02-05
feat: switch over groups

### LINKED ISSUE #182: [Q] Show one group at a time and 'tab' over the groups
It would be nice to be able to treat the groups and the headers like they are tags and be able to tab through them. I am unsure if there is a way I can do this currently out of the box.

Screen Shot 2021-02-03 at 11 58 16 PM

Another time that it would be nice is if I complete a flag or subcommand, and I hit `/` which brings up the fzf UI again, but for files. I know that happens because of the default completions that zsh is doing, but this would be a case where if I got it to do files as one group vs commands vs flags, that tabbing through those groups would be nice.

Aloxaf:
This PR adds a new `tag`: `switch_group` (the name may change later) 

Close #182 

Example usage:
```zsh
# use '-' to go back and '=' to go forward
zstyle ':fzf-tab:*' group-cycle-trigger - =
```


NightMachinary:
How is it that your completions have such fancy groups? The only completion that has groups for me is this example I have defined myself:

```
_foo() {
    A=( -h --help foo bar baz )
    compadd -J group1 -X expl1 -x msg1 -a A
    B=( clown-fish hippo )
    compadd -J group2 -X expl2 -x msg2 -a B
}
function foo { echo $* }
compdef _foo foo
```

image

I have not seen this group feature on any other completion ...

NightMachinary:
I found that you have mentioned in the readme to use `zstyle ':completion:*' format '[%d]'`.

What are those annoying dots that appear at the beginning of completions when groups are enabled?

Aloxaf:
@NightMachinary 

Its added by your configuration. If you're using zsh, please check if you've disabled `COMPLETION_WAITING_DOTS`

NightMachinary:
[**@Aloxaf**](https://github.com/Aloxaf) commented on [Feb 26, 2021, 7:58 PM GMT+3:30](https://github.com/Aloxaf/fzf-tab/pull/183#issuecomment-786751414 "2021-02-26T16:28:59Z - Replied by Github Reply Comments"):
> [@NightMachinary](https://github.com/NightMachinary)
> 
> Its added by your configuration. If you're using zsh, please check if you've disabled `COMPLETION_WAITING_DOTS`

I mean these two dots:
image

Aloxaf:
@NightMachinary Just a prefix.

`zstyle ':fzf-tab:*:' prefix ''`

NightMachinary:
[**@Aloxaf**](https://github.com/Aloxaf) commented on [Feb 26, 2021, 8:07 PM GMT+3:30](https://github.com/Aloxaf/fzf-tab/pull/183#issuecomment-786756300 "2021-02-26T16:37:06Z - Replied by Github Reply Comments"):
> [@NightMachinary](https://github.com/NightMachinary) Just a prefix.
> 
> `zstyle ':fzf-tab:*:' prefix ''`

This is ignored in this line:
```
-ftb-zstyle -s prefix prefix || {
  zstyle -m ':completion:*:descriptions' format '*' && prefix='·'
}
```

To test it, use `open`:
image

(I'm very confused why this prefix is on by default at all. What's its use?)

Aloxaf:
> This is ignored in this line:

Oh sorry I add an extra `:`, it should be `zstyle ':fzf-tab:*' prefix ''`

> (I'm very confused why this prefix is on by default at all. What's its use?)

Help you see the color more clearly.

cldwalker:
I also found the default prefix confusing but not a big deal as it's easy to change. Thanks for the amazing plugin!

kaznovac:
@Aloxaf Switching and filtering by groups is a nice feature for sure, but I can't find how to cancel the group filter (show all completions) once it's activated.
The `.` binding to switch-group is a bit unexpected binding, maybe use `<`, and `>` as they are reserved words (chars) in terminal so it should be safe to assume those chars would not be common in completions

And thanks for the amazing script

cohml:
Is it possible to add an `everything` group that allows for search across the union of all groups? This would be nice in the case that the groups are unfamiliar and it's nonobvious which group would contai the completion I'm looking for.

PR #185, opened 2021-02-05
Fix $ZSH_CUSTOM typo

eliblaney:
There is a small typo in the README.md that says `~ZSH_CUSTOM` instead of `$ZSH_CUSTOM` for the oh-my-zsh installation command.


Aloxaf:
Thanks for your PR.

But `~var` is a valid usage. It will treat (or register if needed) var as a named directory.

eliblaney:
Learned something new today, thanks! Closing this PR.

Issue #184, opened 2021-02-05
[BUG] Parse error for old version of zsh


jrwrigh:
#### Describe the bug
With an old version of zsh, I get a parse error:

```/path/to/plugin/fzf-tab.plugin.zsh:48: parse error near `]]'```

I can make sure:
- [x] I am using the latest version of fzf-tab
- [x] fzf-tab is loaded after `compinit`
- [x] fzf-tab is loaded after plugins which will wrap <kbd>Tab</kbd>, like [junegunn/fzf/completion.zsh](https://github.com/junegunn/fzf/blob/master/shell/completion.zsh)
- [x] fzf-tab is loaded before zsh-autosuggestions, zsh-syntax-highlighting and fast-syntax-highlighting.

#### To Reproduce
1. Source the file

#### Environment:
 - OS:USE Linux Enterprise Server 12 SP5
 - zsh version: 5.0.5

As a minimum working example of the bug:

```zsh
test='test'
if [[ $test ]]; then
    echo 'test worked'
fi
```
This reports the same parsing error as above.

jrwrigh:
I attempted to go through and fix the implicit test (ie. replace `[[ $test ]]` with `[[ -n $test ]]`) which fixes the MWE test I wrote above. However, there appears to still be some other issues related to colors.

First, I get:
```/path/to/repo/fzf-tab/lib/zsh-ls-colors/ls-colors.zsh:114: parse error near `fi'``` 

which is really strange as `fi` doesn't appear anywhere near line 114; the closest are at 124 and 62 (which is in a completely different function).

And then I also get:
```fzf-tab.zsh:339: unknown file attribute: \n```

which is in the definition of `_ftb_group_colors`. Attempting to define the array in a single line returns:
```fzf-tab.zsh:339: unknown file attribute: ^[```

I honestly have no idea why these problems are happening. I've gone ahead and pushed up the commit 
(https://github.com/jrwrigh/fzf-tab/commit/08c2a25bda4618b989b726e0c979d53a76f9b3a7) that has the test expression changes, which would at least be a more robust way of expressing those.

Aloxaf:
I'm sorry, 5.0.5 is too old to support it. I suggest you use https://github.com/romkatv/zsh-bin to get the latest version of zsh in the old distribution.

jrwrigh:
I don't have control over the system. If that was an option, I would've done it already.

jrwrigh:
[...]

PR #186, opened 2021-02-06
refactor: Add explicit non-zero length string tests

jrwrigh:
- Replace implicit `[[ $test ]]` with `[[ -n $test ]]`
 - Doesn't fix the issue reported in #184, but it does at least make the issues more obvious
     - ie. it would've saved me the time in changing all the logical conditions
 - It's also recommended in the [zsh documentation](http://zsh.sourceforge.net/Doc/Release/Conditional-Expressions.html) to use the more explicit version
     - > In other words, [[ $var ]] is the same as [[ -n $var ]]. It is recommended that the second, explicit, form be used where possible.

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -45,15 +45,15 @@
   local key expanded __tmp_value=$'<\0>' # placeholder
   for key in $keys; do
     expanded=${(P)key}
-    if [[ $expanded ]]; then
+    if [[ -n $expanded ]]; then
       __tmp_value+=$'\0'$key$'\0'$expanded
     fi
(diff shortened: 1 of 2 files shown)


Aloxaf:
Thanks for your PR!

But I suggest not to change lib/zsh-ls-colors directly, I suggest sending a PR to https://github.com/xPMo/zsh-ls-colors to avoid conflict in the future.

jrwrigh:
Gotcha. I didn't realize that was a different project. I've removed those changes.

Aloxaf:
LGTM

PR #190, opened 2021-02-16
Fix a bug about globdots

snowphone:
If "globdots" option is enabled, it shows all dotfiles. The problem was, if there's only one or two files in a folder, then fzf-tab only shows only two items and mostly current folder (.) and parent folder (..) consumes these two spaces. It seems showing at least 4 items is an appropriate solution. I actually tried to check whether "globdots" is enabled, like `[[ -o globdots ]]`, but it didn't work.

### DIFF
--- a/lib/-ftb-fzf
+++ b/lib/-ftb-fzf
@@ -45,7 +45,7 @@ print -r -- ${ftb_preview_init/{f}/\$1} > $tmp_dir/ftb_preview_init.$$
 binds=${binds//{_FTB_INIT_}/. $tmp_dir/ftb_preview_init.$$ {f} $'\n'}
 
 local -i header_lines=$#_ftb_headers
-local -i lines=$(( $#_ftb_compcap + 2 + header_lines ))
+local -i lines=$(( $#_ftb_compcap + 4 + header_lines ))
 local reload_command="$commands[zsh] $FZF_TAB_HOME/lib/ftb-switch-group $$ $header_lines $tmp_dir"
 
 # detect if we will use tmux popup
--- a/lib/-ftb-fzf
+++ b/lib/-ftb-fzf
@@ -45,7 +45,11 @@ print -r -- ${ftb_preview_init/{f}/\$1} > $tmp_dir/ftb_preview_init.$$
 binds=${binds//{_FTB_INIT_}/. $tmp_dir/ftb_preview_init.$$ {f} $'\n'}
 
 local -i header_lines=$#_ftb_headers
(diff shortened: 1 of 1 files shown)


NICHOLAS85:
Your issue using `[[ -o globdots]]` was probably because globdots is only responsible for showing dotfiles. `special-dirs` is responsible for showing directory names beginning with `.` You can learn more about the completion option here (scroll down to `special-dirs`): http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Standard-Styles.
You can see what your special-dirs is set to with `zstyle -L ':completion:*' special-dirs`

snowphone:
Now fzf-tab works well with `zstyle -L ':completion:*' special-dirs`. Thanks!

snowphone:
As you said, `zstyle -L ':completion:*' special-dirs` works so I changed the declaration of the variable `lines` in order to show 4 items only if special-dirs is enabled.

Aloxaf:
> The problem was, if there's only one or two files in a folder, then fzf-tab only shows only two items and mostly current folder (.) and parent folder (..) consumes these two spaces. 

I'm sorry. But I can't reproduce your problem. Can you provide clear steps to reproduce it?

> It seems showing at least 4 items is an appropriate solution.

The `+ 2` not means "at least 2 items". It is reserved for the fzf prompt.
图片

snowphone:
My zshrc for reproducing the bug is as follow:
```

unsetopt BEEP	# disable bell
setopt correct
setopt globdots


[ ! -d "${HOME}/.zgen" ] && git clone --depth 1 https://github.com/tarjoilija/zgen.git "${HOME}/.zgen"
source "${HOME}/.zgen/zgen.zsh" > /dev/null

# if the init script doesn't exist
if ! zgen saved; then

  # specify plugins here
  # If zsh-syntax-highlighting is bundled after zsh-history-substring-search,
  # they break, so get the order right.

  zgen oh-my-zsh
  zgen load Aloxaf/fzf-tab

  # generate the init script from plugins above
  zgen save
fi


########################
#  My Configuration    #
########################

#export FZF_DEFAULT_OPTS='-m --border'
#export FZF_DEFAULT_OPTS='-m'
export FZF_DEFAULT_OPTS='--border'
[ -f ~/.fzf.zsh ] && source ~/.fzf.zsh
```
And it seems I found the cause while minimizing my zshrc:  `--border` option. 
Try with, and without `export FZF_DEFAULT_OPTS='--border'` line.

Aloxaf:
I think the best way to solve this problem is to introduce a new tag like `prompt-height`. Because users can also use `--margin` to change the size of the fzf prompt.

PR #197, opened 2021-02-26
reload now runs zsh with -f

NightMachinary:
I have changed the reload comand's `zsh` to `$(which zsh) -f`, as I have a lot of config in my zsh dotfiles which take seconds to load. `which zsh` is also better because `commands[zsh]` mistakenly returns a `zsh` directory in my path (I do have `setopt hash_executables_only`, but I think you do `emulate -L zsh` so it gets voided).

### DIFF
--- a/lib/-ftb-fzf
+++ b/lib/-ftb-fzf
@@ -46,7 +46,7 @@ binds=${binds//{_FTB_INIT_}/. $tmp_dir/ftb_preview_init.$$ {f} $'\n'}
 
 local -i header_lines=$#_ftb_headers
 local -i lines=$(( $#_ftb_compcap + 2 + header_lines ))
-local reload_command="$commands[zsh] $FZF_TAB_HOME/lib/ftb-switch-group $$ $header_lines $tmp_dir"
+local reload_command="$commands[zsh] -f $FZF_TAB_HOME/lib/ftb-switch-group $$ $header_lines $tmp_dir"
 
 # detect if we will use tmux popup
 local use_tmux_popup=0


Aloxaf:
Oh, I forgot to add `-f`. That's a stupid mistake.

But how can `$commands[zsh]` return a path to a directory? Does it return the correct value in the interactive shell?

NightMachinary:
@Aloxaf Without hash_executables_only, no, it returns the wrong path in an interactive session as well. It’s supposedly for performance reasons.

Aloxaf:
@NightMachinary Does this happen with `zsh -f`?

NightMachinary:
@Aloxaf Yes, I’ve asked about it on the mailing list, it’s the intended behavior.

Aloxaf:
Oh, I see. https://inbox.vuxu.org/zsh-workers/CAE9z9A1A-9NFdT2taey6Y18Ek7hg4pRO-goUYvsh-xChriZvGg@mail.gmail.com/t/#u

```zsh
❯ mkdir -p /tmp/test/zsh
❯ PATH=/tmp/test:$PATH
❯ hash -r
❯ echo $commands[zsh]
/tmp/test/zsh
```

But `setopt hash_executables_only` should always work. Can you make sure it is set before any use of `commands` array? Or you can call `hash -r` to rebuild the hash table right after `setopt`.

NightMachinary:
[**@Aloxaf**](https://github.com/Aloxaf) commented on [Feb 27, 2021, 7:01 PM GMT+3:30](https://github.com/Aloxaf/fzf-tab/pull/197#issuecomment-787090402 "2021-02-27T15:31:34Z - Replied by Github Reply Comments"):
> Oh, I see. [inbox.vuxu.org/zsh-workers/CAE9z9A1A-9NFdT2taey6Y18Ek7hg4pRO-goUYvsh-xChriZvGg@mail.gmail.com/t/#u](https://inbox.vuxu.org/zsh-workers/CAE9z9A1A-9NFdT2taey6Y18Ek7hg4pRO-goUYvsh-xChriZvGg@mail.gmail.com/t/#u)
> 
> ```shell
> ❯ mkdir -p /tmp/test/zsh
> ❯ PATH=/tmp/test:$PATH
> ❯ hash -r
> ❯ echo $commands[zsh]
> /tmp/test/zsh
> ```
> 
> But `setopt hash_executables_only` should always work. Can you make sure it is set before any use of `commands` array? Or you can call `hash -r` to rebuild the hash table right after `setopt`.

You're right, doing a `rehash` after setting the option fixed the problem. I have force pushed just the `-f` fix. :D

Issue #206, opened 2021-04-12
[BUG] Typing before fzf has shown up sometimes causes a wrong candidate to be chosen


NightMachinary:
#### Describe the bug

I can make sure:
- [ ] I am using the latest version of fzf-tab
- [ ] this is the minimal zshrc which can reproduce this bug
- [ ] fzf-tab is loaded after `compinit`
- [ ] fzf-tab is loaded after plugins which will wrap <kbd>Tab</kbd>, like [junegunn/fzf/completion.zsh](https://github.com/junegunn/fzf/blob/master/shell/completion.zsh)
- [ ] fzf-tab is loaded before zsh-autosuggestions, zsh-syntax-highlighting and fast-syntax-highlighting.

#### Describe the bug
There is a weird, nondeterministic behavior that is as follows:
1. Press tab on an empty line to trigger the completion of zsh commands, functions, etc. fzf is going to take a while to show up with this much input.
2. Start typing before fzf has shown up.
3. Choose something after fzf appears.
4. SOMETIMES this choice gets replaced with another candidate. When this happens, this wrongly substituted candidate is deterministic (at least for the same characters typed, and possibly regardless of that. It might be the very first candidate available, as it is usually stuff like `!`, `.dumbjump`). 

I had this bug since long ago, but only on macOS, and not on Linux. It used to trigger almost every time. Then I found out that not using `--exit-0` with fzf solves this problem (I had set this option myself, it was not on by default in fzf-tab). But after updating fzf-tab to the "rewrite" that deprecated a bunch of stuff, I now sometimes see this bug. It happens perhaps one out of ten times? I don't know if it happens on Linux or not.

I know this might be hard to get to the bottom of. I just wanted to create an issue so that if other people encounter the same behavior they can also share their perspective.

Aloxaf:
Seems to be related to #175? Does it happen in other terminal emulator in macOS?

NightMachinary:
> Seems to be related to #175? Does it happen in other terminal emulator in macOS?

I thought so too at first, but it’s probably unrelated. The input gets to fzf just fine (the problem is that somehow instead of what is chosen via fzf, the first candidate is outputted), and there is no tmux involved. It happens in both iTerm and Kitty.

NightMachinary:
I tried to debug this, and it seems `fzf` returns the wrong choice:
```
typeset -g -a ch_tmp=( $'ma  \C-@.gdk-pixbuf-post-link.sh\C-@' )

[...]

PR #209, opened 2021-04-19
now empties /dev/tty before passing it to fzf

NightMachinary:
This seems to solve https://github.com/Aloxaf/fzf-tab/issues/206 .

### DIFF
--- a/lib/-ftb-fzf
+++ b/lib/-ftb-fzf
@@ -64,8 +64,15 @@ fi
 
 cat > $tmp_dir/completions.$$
 
+local dd='gdd'
+if (( ${+commands[$dd]} == 0 )) ; then
+  dd='dd'
+fi
+if (( ${+commands[$dd]} == 0 )) ; then
+  dd='true' # nop if dd is not installed
+fi
 
-_ftb_query="${_ftb_query}$(command gdd bs=1G count=1 status=none iflag=nonblock < /dev/tty 2>/dev/null)" || true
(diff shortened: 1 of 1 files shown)


NightMachinary:
Update: Unfortunately, this only somewhat mitigates the problem. `fzf` still behaves weirdly for characters that are entered after `gdd` clears the pipe. I think it's ultimately an fzf bug, though I have not encountered this anywhere else except for `fzf-tab` ...

NightMachinary:
@aloxaf Can you merge this? I recently changed my zsh plugin manager, and so my `fzf-tab` reverted to the latest master, without this patch. The bug appears a lot more often without this patch.

lib/-ftb-fzf
```diff
@@ -64,6 +64,9 @@ fi
 
 cat > $tmp_dir/completions.$$
 
+
+_ftb_query="${_ftb_query}$(command gdd bs=1G count=1 status=none iflag=nonblock < /dev/tty 2>/dev/null)" || true
```
Aloxaf, line 68: I think this only works for macOS users, in Linux there is only `dd`.

You can use `$OSTYPE` to determine which one should be used here.

NightMachinary: @Aloxaf Ok, I added some checks which should be sufficient.

Aloxaf:
LGTM

NightMachinary:
Thanks! <3

PR #216, opened 2021-06-24
Add init.zsh symlink to allow prezto to find the module

AxelTLarsson:
It seems that prezto cannot find the module properly unless there is an init.zsh symlink.


Aloxaf:
Thanks for you PR. But it looks like prezto can load `.plugin.zsh`, have you test it?

https://github.com/sorin-ionescu/prezto/blob/f3c92ffeb3a09acf67bc4c07f56a721a827ac38e/init.zsh#L131

AxelTLarsson:
@Aloxaf Yes, that works too (after updating my prezto to latest)!

PR #251, opened 2021-12-13
Add ability to override fzf keybindings with FZF_TAB_KEYBINDS

kevintraver:
Adding an env variable allows configuration of keybindings specific to `fzf-tab` without having to override or change `fzf` functionality elsewhere.


Aloxaf:
Thanks for you PR.
but you can use `zstyle ':fzf-tab:*' fzf-bindings 'ctrl-j:accept' 'ctrl-a:toggle-all'` instead.

kevintraver:
Ok thanks.

PR #270, opened 2022-03-26
fix: position of tmux popup when status-position is 'top'

ins0mniaque:
This fixes the tmux popup position when the status line is configured to be at the top of the window. It also correctly supports multi-line status.

To test this, here's an example of tmux.conf with 2 lines for the status at the top of the window:
```
set -g status 2
set -g status-position top
```

### DIFF
--- a/lib/ftb-tmux-popup
+++ b/lib/ftb-tmux-popup
@@ -29,7 +29,7 @@ local -a tmp=($(command tmux display-message -p "#{pane_top} #{cursor_y} #{pane_
 local cursor_y=$((tmp[1] + tmp[2])) cursor_x=$((tmp[3] + tmp[4])) window_height=$tmp[5] window_width=$tmp[6]
 
 if [ "$tmp[8]" = 'top' ]; then
-  cursor_y=$((cursor_y + tmp[7] - 1))
+  cursor_y=$((cursor_y + tmp[7]))
 fi
 
 # if not called by fzf-tab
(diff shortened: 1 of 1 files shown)


ins0mniaque:
Sorry for the 'off by 1' fix and then revert.

When testing to add more features, I added a second command called ftb-tmux-popup2, and did not realize the behavior was different because of these lines: https://github.com/Aloxaf/fzf-tab/blob/e8145d541a35d8a03df49fbbeefa50c4a0076bbf/lib/-ftb-fzf#L55-L57

Everything works correctly now.

ins0mniaque:
The last commit fixes the maximum size of the popup so it never shows over the tmux status bar.

lib/ftb-tmux-popup
```diff
@@ -25,8 +25,13 @@ local -a fzf_opts=($@)
 fzf_opts=(${${fzf_opts/--height*}/--layout*})
 
 # get position of cursor and size of window
-local -a tmp=($(command tmux display-message -p "#{pane_top} #{cursor_y} #{pane_left} #{cursor_x} #{window_height} #{window_width}"))
-local cursor_y=$((tmp[1] + tmp[2])) cursor_x=$((tmp[3] + tmp[4])) window_height=$tmp[5] window_width=$tmp[6]
+local -a tmp=($(command tmux display-message -p "#{pane_top} #{cursor_y} #{pane_left} #{cursor_x} #{window_height} #{window_width} #{status} #{status-position}"))
+local cursor_y=$((tmp[1] + tmp[2])) cursor_x=$((tmp[3] + tmp[4])) window_height=$tmp[5] window_width=$tmp[6] window_top=0
+
+if [ "$tmp[8]" = 'top' ]; then
```
Aloxaf, line 31: I suggest use `[[` here to be consistent with project's code style.

Aloxaf:
The rest LGTM, thanks for your contributon!

ins0mniaque:
Like this?

Aloxaf:
Yes. Thanks again!

PR #271, opened 2022-03-31
lib/ftb-tmux-popup: remove awk invocation from comp_length declaration

atomicstack:
Might as well use Perl to calculate the max length too - it might be a micro-optimisation but it's one less `exec()` on every invocation of `ftb-tmux-popup`.


Aloxaf:
LGTM, thanks!

PR #278, opened 2022-04-29
-ftb-fzf: Expand fzf-flags correctly

aeadio:
Word splits `fzf-flags` so that multiple flags can be given


Aloxaf:
Thanks for your contribution.

However, `fzf_flags` is an array, so doesn't need to split it.

Issue #221, opened 2021-08-04
[Q] Disable fzf-tab for a specific command


Coqueiro:
Hi! I really enjoy fzf-tab, but I grew used to how tab completion works for some commands, namely cd. I like the horizontal grid printing the ls. When I want completion, I usually go for my Alt+C. On the other hand, I love how fzf-tab helps me completing things like commands. Is there a way to disable fzf-tab for certain commands?

simon-johansson:
Is there any way of achieving this right now?

PR #293, opened 2022-07-23
feat: toggle fzf-tab for a specific command

badouralix:
Fixes https://github.com/Aloxaf/fzf-tab/issues/221. This pull request introduces a new configuration flag `disabled-on`. It accepts the following values : 

| Value | Description |
|:---:|:--- |
| `any` | Disable fzf-tab for all matches. Fallback to the default completion selection menu. |
| `files` | Only disable fzf-tab if all of the matches are names of files. See flag `-f` in <https://zsh.sourceforge.io/Doc/Release/Completion-Widgets.html#Completion-Builtin-Commands>. |
| `none` | Enable fzf-tab for all matches. |

---

Configuration example for the use-case of https://github.com/Coqueiro

```zsh
# Enable fzf-tab everywhere except on the cd command
zstyle ':fzf-tab:complete:cd:*' disabled-on any
```

Alternative configuration for the use-case of https://github.com/Coqueiro

```zsh
# Enable fzf-tab only to complete non-files, and files are displayed using the regular horizontal grid
zstyle ':fzf-tab:*' disabled-on files
```

Configuration example for the use-case of https://github.com/wookayin

```zsh
# Enable fzf-tab only on specific commands
zstyle ':fzf-tab:*' disabled-on any
zstyle ':fzf-tab:complete:<command in allowlist>:*' disabled-on none
```

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -14,8 +14,14 @@
              i: S: s: I: x: r: R: W: F: M+: E: q e Q n U C \
              J:=__ V:=__ a=__ l=__ k=__ o=__ 1=__ 2=__
 
+  # store $curcontext for further usage
+  _ftb_curcontext=${curcontext#:}
+
   # just delegate and leave if any of -O, -A or -D are given or fzf-tab is not enabled
-  if (( $#_oad != 0 || ! IN_FZF_TAB )); then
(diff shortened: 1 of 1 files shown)


Aloxaf:
Nice implementation. Please update wiki as well.

PR #298, opened 2022-08-06
fix: adjust the ftb-tmux-popup height with info style

georgijd:
When the fzf's --info=inline or --info=hidden flags are passed the
height of the popup should be reduced to match the fzf height

### Before

Screenshot from 2022-08-06 21-38-42

### After
Screenshot from 2022-08-06 21-37-17


Aloxaf:
LGTM

PR #300, opened 2022-08-07
Module build summary

bxsx:
# Abstract

This PR reports a module build summary.

## Pitch

The output of `make` is quite long, and while helpful, it is nice to have an extra one-line summary of the build status. It can save users a lot of time analyzing the logs just to confirm that the module was successfully built and installed.

## Reference
From the `make` manual:
> GNU make exits with a status of zero if all makefiles were successfully parsed and no targets that were built failed.

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -340,6 +340,11 @@ build-fzf-tab-module() {
   pushd $FZF_TAB_HOME/modules
   CPPFLAGS=-I/usr/local/include CFLAGS="-g -Wall -O2" LDFLAGS=-L/usr/local/lib ./configure --disable-gdbm --without-tcsetpgrp ${MACOS:+DL_EXT=bundle}
   make -j$(nproc)
+  if [[ $? -ne 0 ]]; then
+	  echo "The module building has failed. See the output above for details." >&2
+  else
+	  echo "The module has been built sucessfully."
+  fi
   popd
 }
 


fzf-tab.zsh
```diff
@@ -340,6 +340,11 @@ build-fzf-tab-module() {
   pushd $FZF_TAB_HOME/modules
   CPPFLAGS=-I/usr/local/include CFLAGS="-g -Wall -O2" LDFLAGS=-L/usr/local/lib ./configure --disable-gdbm --without-tcsetpgrp ${MACOS:+DL_EXT=bundle}
   make -j$(nproc)
+  if [[ $? -ne 0 ]]; then
```
Aloxaf, line 343: I suggest using `(( ))` here to keep consistent with the rest of the code.

fzf-tab.zsh
```diff
@@ -340,6 +340,11 @@ build-fzf-tab-module() {
...
+	  echo "The module building has failed. See the output above for details." >&2
```
Aloxaf, line 344: I think we can make it more clearly here by printing colored text, like `print -P "%F{red}fail%f %F{green}success%f"`

Aloxaf (changes requested):
This idea is good.

fzf-tab.zsh
bxsx: Good idea!

bxsx:
@Aloxaf, thanks for your feedback! I believe I've addressed all the comments.

Additionally, I moved up `popd`, so it prints the path before the summary line. This has required providing a local variable. If you don't like this solution, alternatively I can redirect the `popd` output to `/dev/null`. Let me know!

Aloxaf (approved):

bxsx:
BTW. How can I test if the module has been built and loaded?

`zmodload` does not output fzftab. I am also wondering why tests fail on macOS?
Regardless of the `make` exit status, I am confused if the whole process runs successfully ;)

PR #301, opened 2022-08-09
fix: CPU cores info on macOS

bxsx:
# Problem

`nproc` does not exist in macOS, therefore the number of CPU cores is not taken, nor additional cores are not utilized on macOS during the `build-fzf-tab-module()` call.

# Solution

This PR solves this problem by using `sysctl` on macOS systems.


Aloxaf:
LGTM

PR #324, opened 2022-11-03
fix: Only add to `$FPATH` once

dmorgan-fa:
# What
Rather than adding to `$FPATH` every time a new shell is created, check to see if it's already present. This means we don't end up with multiple instances of `${FZF_TAB_HOME}/lib` included in `$FPATH`.

# Example
## Before
```
$ echo $FPATH | tr ":" "\n"
[...]
${HOME}/fzf-tab/lib
$ zsh   # open a subshell
$ echo $FPATH | tr ":" "\n"
[...]
${HOME}/fzf-tab/lib
${HOME}/fzf-tab/lib
```

## After
```
$ echo $FPATH | tr ":" "\n"
[...]
${HOME}/fzf-tab/lib
$ zsh   # open a subshell
$ echo $FPATH | tr ":" "\n"
[...]
${HOME}/fzf-tab/lib
```

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -379,7 +379,9 @@ typeset -ga _ftb_group_colors=(
 () {
   emulate -L zsh -o extended_glob
 
-  fpath+=($FZF_TAB_HOME/lib)
+  if [[ ! "$FPATH" == *${FZF_TAB_HOME}/lib* ]]; then
+    fpath+=($FZF_TAB_HOME/lib)
+  fi
 
   autoload -Uz is-at-least -- $FZF_TAB_HOME/lib/-#ftb*(:t)
 
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -379,7 +379,7 @@ typeset -ga _ftb_group_colors=(
 () {
   emulate -L zsh -o extended_glob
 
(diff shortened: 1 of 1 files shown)


Aloxaf:
Sorry, I cannot reproduce this. What's your zsh version?

图片

dmorgan-fa:
Ah, I see it's only happening on my Mac:
```
zsh 5.8.1 (x86_64-apple-darwin22.0)
```

I get the same experience on Ubuntu as you, without the duplicate entries in $fpath:
```
zsh 5.8 (x86_64-ubuntu-linux-gnu)
```

fzf-tab.zsh
```diff
@@ -376,7 +376,9 @@ typeset -ga _ftb_group_colors=(
 () {
   emulate -L zsh -o extended_glob
 
-  fpath+=($FZF_TAB_HOME/lib)
+  if [[ ! "$FPATH" == *${FZF_TAB_HOME}/lib* ]]; then
```
Aloxaf, line 379:
```suggestion
  if (( ! $fpath[(I)$FZF_TAB_HOME/lib] )); then
```

dmorgan-fa: That's a much more specific check, nice solution 👌

Aloxaf:
Thanks for your contribution!

PR #326, opened 2022-11-07
feat: swap tab and btab direction when tmux-popup shows above the cursor

zyf0330:
Do you think this is a good change?

### DIFF
--- a/lib/ftb-tmux-popup
+++ b/lib/ftb-tmux-popup
@@ -70,6 +70,7 @@ if (( cursor_y * 2 > window_height )); then
   # show above the cursor
   popup_height=$(( min(max(comp_lines + 4, popup_min_size[2]), cursor_y - window_top) + adjust_height ))
   popup_y=$cursor_y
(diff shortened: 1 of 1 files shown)


Aloxaf:
Quite reasonable.

tardypad:
I think this commit should be reverted because it doesn't follow the usual behavior of fzf.
fzf doesn't invert the meaning of `up` and `down` actions depending on the layout.
So this brings inconsistencies in the usage of fzf.
For example when used in fzf-tab with a default layout `tab` goes up visually but in other instances, it would go down.

Personnally this change broke the usage of fzf-tab because I force fzf to use reverse layout by default.

Aloxaf:
@tardypad 
Thanks for your advice. I add a option named `popup-smart-tab` to control this behavior.
You can use `zstyle ':fzf-tab:complete:*' popup-smart-tab false` to disable it now.

tardypad:
@Aloxaf Thanks for the really quick change.
I can live with this workaround :)

PR #329, opened 2022-11-26
Update config.guess and config.sub - previous versions fail on Raspbian OS

NodeJSmith:
Not sure if there are other considerations before updating these, but the previous version are from 2009 while the most recent version is from 2022. The updated versions recognize Raspian while the previous did not, which is why I ended up here.


Aloxaf:
I update the build script to pull zsh source code from github. Can you try https://github.com/Aloxaf/fzf-tab/pull/334?

PR #351, opened 2023-02-15
feat: close fzf selection if backspace is pressed on empty input

qoomon:


qoomon:
@Aloxaf any feedback?

yrammos:
@qoomon I think this setting pertains to `fzf`, not `fzf-tab`. Have you tried binding backspace to the `backward-delete-char/eof` action?

`fzf --bind=bspace:backward-delete-char/eof`

qoomon:
> Have you tried binding backspace to the backward-delete-char/eof action?

@yrammos that's exactly what I did in this PR

yrammos:
Oops, indeed -- sorry!

Aloxaf:
That's good! 

But I suggest moving it here, so that user can override it if they want:
https://github.com/Aloxaf/fzf-tab/blob/f95cdcf2c390428d271f2304698c2d45f0cd4de5/lib/-ftb-fzf#L31

qoomon:
@Aloxaf done.

PR #363, opened 2023-03-30
tmux-popup: conditionally import math functions

avamsi:
Without this, I'm seeing "min / max / sum" already defined errors on
RHEL 8, the previous definition presumably from -ftb-fzf.

This seems independent of the Zsh version and I'm not really sure what
changed between RHEL 7 and RHEL 8 that triggers this.


avamsi:
@Aloxaf, would you be open to merging this?

Aloxaf:
LGTM, thanks for your contribution!

PR #421, opened 2024-03-07
Fix path collision on case-insensitive FS

### LINKED ISSUE #420: [BUG] source directory name confliction on case-insensitive filessytem such as APFS for  macos
#### Describe the bug

`modules/src` and `modules/Src` are different folders on case-sensitive filesystem such as ext4, btrfs and so on. 

But on case-insensitive filesystem, such as APFS for macos, the one of them will be missing. This causes that configure and build process fail.

dbaynard:
The `modules/Src` and `modules/src` paths collide on MacOS (for example) and `Src` is preferred, meaning builds fail.

This replaces the `modules/Src/Aloxaf.gitignore` with entries in the `modules/.gitignore` that does a similar thing (it actually ignores the whole `Src` directory).

Fixes #420


Aloxaf:
This is a historical issue. The `Src` directory was used to store built module. Removing it could cause build process failures or users have to to rebuild the module. 

I suggest merging the `src` directory into `Src`.

Relevant code:

https://github.com/Aloxaf/fzf-tab/blob/f3d892cf19bba09c0c44fa9fcaf6d616fe5b18e6/lib/-ftb-build-module#L18-L19

doubleloop:
In my case, even though I use linux, I keep dotfiles within Dropbox which does not allow for case collisions within synced folder so currently I can not checkout master without issues...

PR #426, opened 2024-04-02
Reset prompt after completion done

hotpxl:
Somehow if I don't call `zle reset-prompt`, the highlighting turns off when the completion is done (either there's no more choice or when I ctrl-c)

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -256,6 +256,7 @@ fzf-tab-complete() {
     fi
   done
   echoti cnorm >/dev/tty 2>/dev/null
+  zle reset-prompt
   zle .redisplay
   (( _ftb_accept )) && zle .accept-line
   return $ret


hotpxl:
Actually, it also works if I call `zle redisplay` instead of `zle .redisplay`.
Is there any reason we are calling the original `redisplay` function? In my case, I use z-sy-h and it overrides widgets to highlight stuff.

hotpxl:
@Aloxaf any comment is much appreciated!

Aloxaf:
What is your plugin loading order? Did you make sure fzf-tab was loaded before z-sy-h?

hotpxl:
yeah, fzf-tab is loaded before z-sy-h

Aloxaf:
what's the output of `echo $widgets[fzf-tab-complete]`?

hotpxl:
```
echo $widgets[fzf-tab-complete]
user:_zsh_autosuggest_bound_1_fzf-tab-complete
```

Aloxaf:
can you remove zsh-autosuggestions and try again?

hotpxl:
hi, shall we merge this?

PR #438, opened 2024-04-27
Introduce `fzf-layout` setting

sobolevn:
Hi! Thanks a lot for the amazing library, I want to customize the layout of tab completions.

Right now I cannot do that with `zstyle ':fzf-tab:*' fzf-flags '--layout=reverse-list'`, because `fzf` does not allow two `--layout` flags at the same time.

Can we please make this customizable? :)


orsinium (approved):

Aloxaf:
> Right now I cannot do that with zstyle ':fzf-tab:*' fzf-flags '--layout=reverse-list', because fzf does not allow two --layout flags at the same time.

what's your fzf version? I test it with 0.48.1 and it works.

sobolevn:
Indeed, the update worked! Thanks!

PR #445, opened 2024-05-21
feat(fit-preview): optionally double comp_length if when showing preview

jimeh:
Note: This was a 5 minute hack just to see if I could fix it, so more than happy to hear any feedback and change things as needed :)

---

When previews are enabled, they typically take up half the width of the popup. This adds an option to make it double the width when preview is enabled, hence ensuring that the full text of completion items is visible.

### Before

With preview: 

Screen-Shot-2024-05-21-01-42-29 03

Without preview:

Screen-Shot-2024-05-21-01-48-16 96

### After

With preview:

Screen-Shot-2024-05-21-01-43-04 51

Without preview:

Screen-Shot-2024-05-21-01-47-20 60


z0rc:
@Aloxaf ping. Currently `ftb-tmux-popup` isn't usable together with `fzf-preview`.

Sneethe:
Very cool. This solves my main gripe with fzf-tab.

But I wonder, would it be difficult to trigger resizing the popup pane when toggling the preview?

I use [fzf-tab-source](https://github.com/Freed-Wu/fzf-tab-source) for rich preview windows, but they're more something you want to toggle on when needed rather than having them always on. Meaning the pane will always be double its necessary width, just for the off chance the preview is toggled on.

Sneethe:
Very **crude** POC 
```
tmux display-popup -E -w 40% -h 70% -x 10% -y 10% "fzf" &
sleep 5                 
kill $(pgrep -n "fzf")
sleep .1 
tmux display-popup -E -w 80% -h 70% -x 10% -y 10% "fzf"
```

lib/ftb-tmux-popup
```diff
@@ -45,6 +45,10 @@ local text REPLY comp_lines comp_length length popup_pad popup_min_size
 
 zstyle -a ":fzf-tab:$_ftb_curcontext" popup-pad popup_pad || popup_pad=(0 0)
 zstyle -a ":fzf-tab:$_ftb_curcontext" popup-min-size popup_min_size || popup_min_size=(0 0)
+zstyle -a ":fzf-tab:$_ftb_curcontext" popup-fit-preview popup_fit_preview_enabled || popup_fit_preview_enabled=no
+if [ "$popup_fit_preview_enabled" != "no" ] && zstyle -m ":fzf-tab:$_ftb_curcontext" fzf-preview '*'; then
+  popup_fit_preview=1
+fi
```
Sneethe, line 51: This.

Sneethe (commented):
can be simplified down to just:
```
if zstyle -t ":fzf-tab:$_ftb_curcontext" popup-fit-preview && zstyle -m ":fzf-tab:$_ftb_curcontext" fzf-preview '*'; then
  comp_length=$(( comp_length * 2 ))
fi

```

Issue #475, opened 2024-10-02
[Q] Using `FZF_DEFAULT_OPTS`.


pablospe:
After this commit: 0b49f3e8a3fdb546636ff2157925d32e8936b415 (`fix: always ignore FZF_DEFAULT_OPTS`)
How do you set to actually use `FZF_DEFAULT_OPTS`, if one wants to?

PrayagS:
+1. Dumping all my flags in `fzf-flags` has no effect for some reason.

Can this be turned into an optional behavior? Default can be to ignore `FZF_DEFAULT_OPTS`. Happy to open a PR if that sounds good.

pablospe:
> Can this be turned into an optional behavior? Default can be to ignore `FZF_DEFAULT_OPTS`. Happy to open a PR if that sounds good

If you know how to do it, please create a PR. I think few people will benefit from it, even if it is not merged in main. I understand the reason why to ignore `FZF_DEFAULT_OPTS`, but it also creates a usability problem (at least it is not document properly, as far as I understand, perhaps it is an easy fix).

PrayagS:
> If you know how to do it, please create a PR. I think few people will benefit from it, even if it is not merged in main.

Opened the PR: https://github.com/Aloxaf/fzf-tab/pull/479

> I understand the reason why to ignore FZF_DEFAULT_OPTS, but it also creates a usability problem (at least it is not document properly, as far as I understand, perhaps it is an easy fix).

It was documented recently from what I see. https://github.com/Aloxaf/fzf-tab/blob/cf57116eb2410139b8bd78fcb793dd84db79e28f/README.md?plain=1#L118-L119

The issue I saw with using `fzf-flags` is that most users of this plugin also use https://github.com/Freed-Wu/fzf-tab-source/ which overrides the value of `fzf-flags`. The final value of `fzf-flags` that I was seeing in my shell wasn't showing up as expected.

If someone is not using that plugin, I suppose `fzf-flags` should work directly without needing this patch.

Aloxaf:
You can use `FZF_DEFAULT_OPTS` by this:

```zsh
zstyle ':fzf-tab:*' fzf-flags ${(z)FZF_DEFAULT_OPTS}
```

pablospe:
Thanks for the answer, @Aloxaf! Unfortunately, this didn't work for me, see this comment:
https://github.com/Aloxaf/fzf-tab/pull/479#discussion_r1791920144

PR #479, opened 2024-10-06
Add `use-fzf-default-opts`

PrayagS:
Ref: https://github.com/Aloxaf/fzf-tab/issues/475

https://github.com/Aloxaf/fzf-tab/commit/0b49f3e8a3fdb546636ff2157925d32e8936b415 introduces a change that makes the plugin ignore `FZF_DEFAULT_OPTS` by default.

This commit adds a flag to disable this behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,10 @@ zstyle ':completion:*' menu no
 # preview directory's content with eza when completing cd
 zstyle ':fzf-tab:complete:cd:*' fzf-preview 'eza -1 --color=always $realpath'
 # custom fzf flags
-# NOTE: fzf-tab does not follow FZF_DEFAULT_OPTS
+# NOTE: fzf-tab does not follow FZF_DEFAULT_OPTS by default
 zstyle ':fzf-tab:*' fzf-flags --color=fg:1,fg+:2 --bind=tab:accept
+# To make fzf-tab follow FZF_DEFAULT_OPTS
+zstyle ':fzf-tab:*' use-fzf-default-opts yes
 # switch group using `<` and `>`
 zstyle ':fzf-tab:*' switch-group '<' '>'
 ```
--- a/lib/-ftb-fzf
+++ b/lib/-ftb-fzf
@@ -40,6 +40,7 @@ local ret=0
 -ftb-zstyle -a switch-group switch_group || switch_group=(F1 F2)
 -ftb-zstyle -s fzf-pad fzf_pad || fzf_pad=2
 -ftb-zstyle -s fzf-min-height fzf_min_height || fzf_min_height=0
+-ftb-zstyle -b use-fzf-default-opts use_fzf_default_opts || use_fzf_default_opts="no"
 
 -ftb-zstyle -a debug-command debug_command && {
   ${(eX)debug_command} $fzf_flags
@@ -81,7 +82,12 @@ fi
 
 _ftb_query="${_ftb_query}$(command "$dd" bs=1G count=1 status=none iflag=nonblock < /dev/tty 2>/dev/null)" || true
 
-FZF_DEFAULT_OPTS='' SHELL=$ZSH_NAME $fzf_command \
+local fzf_default_opts=''
+if [[ "$use_fzf_default_opts" == "yes" ]]; then
+  fzf_default_opts=$FZF_DEFAULT_OPTS
+fi
+
+FZF_DEFAULT_OPTS=$fzf_default_opts SHELL=$ZSH_NAME $fzf_command \
   --ansi \
   --bind=$binds \
   --bind="${switch_group[1]}:reload($reload_command -1),${switch_group[2]}:reload($reload_command 1)" \


pablospe (approved):

lib/-ftb-fzf
```diff
@@ -81,7 +82,12 @@ fi
 
 _ftb_query="${_ftb_query}$(command "$dd" bs=1G count=1 status=none iflag=nonblock < /dev/tty 2>/dev/null)" || true
 
-FZF_DEFAULT_OPTS='' SHELL=$ZSH_NAME $fzf_command \
+local fzf_default_opts=''
+if [[ "$use_fzf_default_opts" == "yes" ]]; then
+  fzf_default_opts=$FZF_DEFAULT_OPTS
```
pablospe, line 87: I was wondering if `fzf_default_opts` couldn't be set outside like this:
`zstyle ':fzf-tab:*' fzf-default-opts $FZF_DEFAULT_OPTS`
?

I that case you could get ride off the `use_fzf_default_opts` option.

pablospe:
Adding @Aloxaf for awareness. I think it is good to add an option so the user can decide about this behavior but taking care of leaving a safe default, as this PR proposes.

lib/-ftb-fzf
PrayagS: Yeah I was going for a quick change so didn't try that. I thought `zstyle` might parse things differently and lead to unintended behaviors.

Let me try that and get back here.

PrayagS: Ah I remember it now. Since `FZF_DEFAULT_OPTS` is a bunch of flags, we need to pass it as such and not as an environment variable.

And `fzf-flags` already exists in the plugin. But when I tried setting that equal to the flags in `FZF_DEFAULT_OPTS` (see below), it wasn't working. I think the reason is the same as what I mentioned in https://github.com/Aloxaf/fzf-tab/issues/475#issuecomment-2395474675.

```sh
# this converts the string to a string array which is expected by fzf-flags.
export parsed_fzf_default_opts=(${=FZF_DEFAULT_OPTS})

zstyle ':fzf-tab:*' fzf-flags $parsed_fzf_default_opts
```

pablospe: Your suggestion doesn't seem to be working for my case:
```
export FZF_DEFAULT_OPTS="          \
  --exact                          \
  --height ${FZF_TMUX_HEIGHT:-30%} \
  --layout=reverse                 \
  --info=hidden                    \
  --prompt='$  '                   \
  --pointer='=>'                   \
  --marker='+'                     \
  --cycle                          \
  --select-1                       \
  --color=dark                     \
  --bind '?:toggle-preview,ctrl-a:select-all,ctrl-d:half-page-down,ctrl-u:half-page-up,ctrl-l:clear-query,home:first,end:last,enter:accept-non-empty,alt-down:down,alt-up:up,right:accept-non-empty,alt-right:accept-non-empty,alt-left:close,left:close' \
  --color='fg:-1,bg:-1,hl:#fb8aa4,fg+:-1,bg+:-1,hl+:#55E579' \
  --color='info:#af87ff,prompt:#fb8aa4,pointer:#55E579,marker:#55E579,spinner:#55E579'
"
```

I think the `--height ${FZF_TMUX_HEIGHT:-30%}` is not correctly handle.

Aloxaf: @pablospe can you try this?

```zsh
zstyle ':fzf-tab:*' fzf-flags ${(Q)${(Z:n:)FZF_DEFAULT_OPTS}}
```

README.md
```diff
@@ -115,8 +115,10 @@ zstyle ':completion:*' menu no
 # preview directory's content with eza when completing cd
 zstyle ':fzf-tab:complete:cd:*' fzf-preview 'eza -1 --color=always $realpath'
 # custom fzf flags
-# NOTE: fzf-tab does not follow FZF_DEFAULT_OPTS
+# NOTE: fzf-tab does not follow FZF_DEFAULT_OPTS by default
 zstyle ':fzf-tab:*' fzf-flags --color=fg:1,fg+:2 --bind=tab:accept
+# To make fzf-tab follow FZF_DEFAULT_OPTS
```
Aloxaf, line 120: I think we should warn user that this may cause unexpected behavior.

Aloxaf:
Umm, quite reasonable.

`zstyle ':fzf-tab:*' fzf-flags ${(Q)${(Z:nC:)FZF_DEFAULT_OPTS}}` or `zstyle ':fzf-tab:*' use-fzf-default-opts yes`, which one do you prefer?

lib/-ftb-fzf
pablospe: This is magic! It is working, thanks!

PrayagS: These are my flags,
```
# Set fzf options
export FZF_DEFAULT_OPTS=$FZF_DEFAULT_OPTS'
    --color=bg:#1b1b1b
    --color=preview-bg:#1b1b1b
    --color=hl+:#fb4934
    --color=gutter:#1b1b1b
    --color=pointer:#f9f5d7,marker:#f9f5d7
    --color=info:#ebdbb2,spinner:#f9f5d7
    --color=query:#ebdbb2,prompt:#ebdbb2
    --color=border:#ebdbb2
    --border="rounded" --prompt="> "
    --marker="" --pointer="->" --layout="reverse"
'
```

While `echo ${(Q)${(Z:nC:)FZF_DEFAULT_OPTS}}` looks as expected, when I try to open fzf-tab, it fails with an eval error.

pablospe:
> Umm, quite reasonable.
> 
> `zstyle ':fzf-tab:*' fzf-flags ${(Q)${(Z:nC:)FZF_DEFAULT_OPTS}}` or `zstyle ':fzf-tab:*' use-fzf-default-opts yes`, which one do you prefer?

I don't have a strong preference:
`zstyle ':fzf-tab:*' fzf-flags ${(Q)${(Z:nC:)FZF_DEFAULT_OPTS}}` 
doesn't touch the code, only the documentation. Although, it is more complicated to understand, and my guess it is more suceptible to errors (maybe something in the FZF_DEFAULT_OPTS will break the parsing). In the other hand, the 2nd option seems more elegant, there is no parsing and self-explanatory: `use-fzf-default-opts yes`. Notice that implementing the 2nd solution still allows to do use the first one.

README.md
pablospe: @PrayagS I think you can mention here there are other plugins that overrides the value of fzf-flags directly, creating bugs.

Now I was wondering if overriding the fzf-flags is the right thing to do for an external plugin, it should do something like: fzf-flags = fzf-flags + <extra_things>?

PrayagS: @Aloxaf Updated the README. Please review [`380daee` (#479)](https://github.com/Aloxaf/fzf-tab/pull/479/commits/380daeeb8af1972b901dc873c4fe4a74f897bb21).

@pablospe My comment on Freed-Wu/fzf-tab-source causing a conflict is a hunch. Because when I was reviewing its code, it overrides it specifically for a completion type (see https://github.com/Freed-Wu/fzf-tab-source/blob/aabde06d1e82b839a350a8a1f5f5df3d069748fc/fzf-tab-source.plugin.zsh#L31). The more likely reason for it to not work could be just a parsing issue with the flags.

pablospe (approved):

Issue #477, opened 2024-10-03
[BUG] --color=fg:*,fg+:* is not respected, but works in fzf


egirlcatnip:
#### Describe the bug

I can make sure:
- [x] I am using the latest version of fzf-tab
- [x] this is the minimal zshrc which can reproduce this bug
- [x] fzf-tab is loaded after `compinit`
- [x] fzf-tab is loaded after plugins which will wrap <kbd>Tab</kbd>, like [junegunn/fzf/completion.zsh](https://github.com/junegunn/fzf/blob/master/shell/completion.zsh)
- [x] fzf-tab is loaded before zsh-autosuggestions, zsh-syntax-highlighting and fast-syntax-highlighting.

#### To Reproduce
1. set ``zstyle ':fzf-tab:*' fzf-flags --color=fg:5``
2. compare to command ``fzf --color=fg:5``

#### Expected behavior
fg and fg+ color is respected in the configuration. This way, in gnome's light mode, the foreground is hard to read.

#### Screenshots
Bug:
![image](https://github.com/user-attachments/assets/90fc4195-692e-46ce-ae64-255bf1efce97)
Expected:
![image](https://github.com/user-attachments/assets/b3775555-cc01-4c4a-b08b-4cc12ab2920a)

#### Environment:
 - OS: Fedora 40
 - zsh version: 5.9

zhermin:
Same here, tried a few combinations of `zstyle` flags and location in `.zshrc` and following the README but the colors don't persist.

Aloxaf:
@pivoshenko your problem is a duplicate of #475

Aloxaf:
This is strange, I cannot reproduce this. fzf-flags is appended to the end of the fzf commandline, it should work. 

Can you make sure you only set `fzf-flags` once? And please provide the log (trigger fzf-tab with `Ctrl-x .`).

pablospe:
> ```shell
> --color=bg+:#363a4f,bg:#24273a,spinner:#f4dbd6,hl:#ed8796
> ```

Could you try adding `'` (single quote) like this:
`--color='bg+:#363a4f,bg:#24273a,spinner:#f4dbd6,hl:#ed8796'`

peterldowns:
@egirlcatnip I was having this issue as well — no matter what I did, the foreground text color was ALWAYS "white". After messing around debugging in the source of fzf-tab, I believe the issue is due to the default "group color":

[...]

PR #484, opened 2024-11-05
Remove text-coloring that interferes with fzf theming

peterldowns:
This PR is based on the thread of discussion in https://github.com/Aloxaf/fzf-tab/issues/477. The goal of this PR is to make it so that in order to theme fzf-tab with the same colors as your normal fzf theme, users can use just the following settings in their .zshrc:

```zsh
zstyle ':fzf-tab:*' use-fzf-default-opts yes
source ~/code/fzf-tab/fzf-tab.zsh
```

As fzf-tab works today, the required settings are:

```zsh
zstyle ':fzf-tab:*' use-fzf-default-opts yes
# necessary to override the hard-coded color flag passed to fzf by fzf-tab:
#
#     --color=hl:$(( header_lines == 0 ? 188 : 255 )) \ 
#
zstyle ':fzf-tab:*' fzf-flags --color=hl:myThemeColor
# necessary to replace the default value of \[37m, "hard white",
# which is used when there is only one completion group — which
# overrides any color values of `hl` or `hl+` being set in other fzf flags
#
#      -ftb-zstyle -s default-color default_color || default_color=$'\x1b[37m' 
#
zstyle ':fzf-tab:*' default-color ""
source ~/code/fzf-tab/fzf-tab.plugin.zsh
```

As a new user of fzf-tab, I am extremely grateful for this project. It works very well and between the fzf knowledge, shell knowledge, and c knowledge required to make it work so quickly, I am seriously impressed by the work that has gone into it. I am hoping that this change will make the user experience better by making it way easier to theme fzf-tab to look like any other fzf instance.

This would be a breaking change. I do not know the history of these default values, but the single-group default color has been \[37m since grouping was introduced, although it was originally known as `FZF_TAB_NO_GROUP_COLOR`: https://github.com/Aloxaf/fzf-tab/pull/23/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R121

I generally believe that PRs to update default themes/colors should be rejected by maintainers by default, but in this case I'd like you to consider that these changes only make it easier to customize fzf-tab's theme and make it look the way the user wants, by removing defaults present in the project.

I tested this change by using these zshrc settings, and observing that the fzf style matched my default fzf theme as expected, in both light mode and in dark mode:

```zsh
zstyle ':fzf-tab:*' use-fzf-default-opts yes
source ~/code/fzf-tab/fzf-tab.plugin.zsh
```

If this PR is accepted as-is, [the wiki entry for `default-color`](https://github.com/Aloxaf/fzf-tab/wiki/Configuration#default-color) will also need to be updated.

### DIFF
--- a/lib/-ftb-fzf
+++ b/lib/-ftb-fzf
@@ -91,7 +91,6 @@ FZF_DEFAULT_OPTS=$fzf_default_opts SHELL=$ZSH_NAME $fzf_command \
   --ansi \
   --bind=$binds \
   --bind="${switch_group[1]}:reload($reload_command -1),${switch_group[2]}:reload($reload_command 1)" \
-  --color=hl:$(( header_lines == 0 ? 188 : 255 )) \
   --cycle \
   --delimiter='\x00' \
   --expect=$continuous_trigger,$print_query,$accept_line \
--- a/lib/-ftb-generate-complist
+++ b/lib/-ftb-generate-complist
@@ -9,7 +9,7 @@ local -A word_map=()
 (( $#_ftb_compcap == 0 )) && return
 
 -ftb-zstyle -s show-group show_group || show_group=full
--ftb-zstyle -s default-color default_color || default_color=$'\x1b[37m'
+-ftb-zstyle -s default-color default_color || default_color=''
 -ftb-zstyle -s prefix prefix || {
   zstyle -m ':completion:*:descriptions' format '*' && prefix='·'
 }


villem:
This pull request fixes the hl+ coloring issues. I can now use any zsh-tinted theme. Excellent! And +1 fzf-tab being superb, utility.

peterldowns:
@Aloxaf thank you for merging my contribution, and thank you for creating and maintaining fzf-tab.

Now that this PR is merged, you should probably update [the wiki entry for `default-color`](https://github.com/Aloxaf/fzf-tab/wiki/Configuration#default-color) and close issue #477.

Thanks again 🙇

Aloxaf:
Done.
Thanks for your contribution.

Issue #394, opened 2023-10-14
[FR] add aliased command to `$desc`


oldfatcrab:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when using the `git` alias, there are too many `g...` alias I can choose (i.e. `alias gpsup="git push --set-upstream origin $(git_current_branch)"`).

**Describe the solution you'd like**
If the aliased command can be written into `$desc` that will be great. (and so on for shell function and parameter)

**Describe alternatives you've considered**
Or some documentation on how to customize `$desc` field will be also good.

**Additional context**
![image](https://github.com/Aloxaf/fzf-tab/assets/1563274/8a0c6141-8ada-4976-8a36-04fee4a06ee9)

PR #499, opened 2024-12-25
More descriptive alias

oldfatcrab:
This is my quick and dirty way to solve the problem mentioned in https://github.com/Aloxaf/fzf-tab/issues/394

Not even certain if this is the correct approach (not very efficient to me) - feel free to close mine and make another PR if a better approach exists.

Short demo: https://youtu.be/L-Yml_kN9eU

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -80,6 +80,8 @@ builtin unalias -m '[^+]*'
       word=$__hits[i] dscr=$__dscr[i]
       if [[ -n $dscr ]]; then
         dscr=${dscr//$'\n'}
+      elif [[ "$expl" == '[alias]' ]]; then 
+        dscr="$(echo $(alias $word) | sed 's/=/  :=>  /')"
       elif [[ -n $word ]]; then
         dscr=$word
       fi


Sneethe:
This is a really neat feature.
It however doesn't work when [fzf-tab-source](https://github.com/Freed-Wu/fzf-tab-source) is used.

Can be tested with the following:
```zsh
cd /tmp
cd "$(mktemp -d -- "${PWD}/tmp.XXXXXXXXXX")"
git clone https://github.com/oldfatcrab/fzf-tab
git clone https://github.com/Freed-Wu/fzf-tab-source

> minzshrc << 'EOF'
autoload -Uz compinit
compinit
zstyle ':completion:*:descriptions' format '[%d]'
alias tree='tree -a -I .git'
source fzf-tab/fzf-tab.plugin.zsh
source fzf-tab-source/fzf-tab-source.plugin.zsh
EOF
```
Then just run `zsh -d -f`
and then:
`source minzshrc`

`tree` will only show its alias content when fzf-tab-source is commented

oldfatcrab:
@Sneethe at first I thought it was my branch is too old but it wasn't caused by that.

I am not even certain that my change is proper because the searching is not (that) efficient.

My recommendation is to check how `zstyle ':fzf-tab:complete:*:*' fzf-preview ` is set in [fzf-tab-source](https://github.com/Freed-Wu/fzf-tab-source).

On my machine I set it like [this](https://github.com/oldfatcrab/.dotfiles/blob/83e9df56e7d453474c14842e5a8ae963127a2ea0/zsh/zshrc.zsh#L141)

So the preview for directories works for me like:
Screenshot 2025-06-15 at 01 22 05

oldfatcrab:
@Sneethe Actually I just (briefly) went through the code of [fzf-tab-source](https://github.com/Freed-Wu/fzf-tab-source), and I think it is better for you to submit a feature request - I think a code change on that side could yield more sufficient result.

PR #514, opened 2025-03-16
Avoid spawning a subshell in `-ftb-compadd`

### LINKED ISSUE #513: [BUG] `$PROMPT` is not quoted correctly when expanded
#### Describe the bug

If I have a `PROMPT` that runs a shell command, the quotes in it are not preserved properly when running `fzf-tab-complete`.

I can make sure:

#### To Reproduce
1. Launch a zsh shell without any configuration: `env -i HOME="$HOME" TERM="$TERM" PS1='$ ' zsh --no-rcs --no-globalrcs`
2. Source the following script:
```sh
autoload -Uz compinit
compinit
source fzf-tab/fzf-tab.zsh

setopt PROMPT_SUBST
foo() {
	# set -x
	echo hi | grep ^"xy" >/dev/null
	echo '$ '
}
PROMPT='$(foo)'
```
2. Type <kbd>ls&nbsp;</kbd> (with the trailing space)
3. Press <kbd>Tab</kbd>
4. See this output:
```
$ ls grep: bin: Is a directory
grep: config: Is a directory
grep: lib: Is a directory
```

If you enable `set -x` in the script, you can see that grep is being called with all the completions as arguments:
```
$ ls +foo:2> echo hi
+foo:2> grep LICENSE README.md bin config config.txt defaults header.tex lib packages.txt python.txt replicate.sh rust.txt save-me setup.ps1 setup.sh signature.html tmp track.sh
```

This is specifically about quoting; if I change the second line of `foo` to `echo hi | grep "^xy" >/dev/null` (notice the new double-quotes) then this all works fine.

#### Expected behavior

The function called in PROMPT should be quoted correctly.

#### Screenshots

Image

#### Environment:
 - OS: Linux pop-os 6.9.3-76060903-generic
 - zsh version: zsh 5.8.1 (x86_64-ubuntu-linux-gnu)

#### Minimal zshrc

See above.

#### Log

[zsh-3810555-fzf-tab-7.txt](https://github.com/user-attachments/files/19268929/zsh-3810555-fzf-tab-7.txt)

Notice in particular line 724, which has the incorrectly quoted line.

jyn514:
Subshells cause zsh to try and render their prompt. For reasons I do not entirely understand, zsh does not expand `$PROMPT` correctly in this case when PROMPT_SUBST is set.

Fortunately, using a brace group instead of a subshell is simple and has the same meaning.

Fixes #513

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -25,7 +25,7 @@ builtin unalias -m '[^+]*'
   # or fzf-tab is disabled in the current context
   if (( $#_oad != 0 || ! IN_FZF_TAB )) \
     || { -ftb-zstyle -m disabled-on "any" } \
-    || ({ -ftb-zstyle -m disabled-on "files" } && [[ -n $isfile ]]); then
+    || { -ftb-zstyle -m disabled-on "files" && [[ -n $isfile ]] }; then
     builtin compadd "$@"
     return
   fi


Aloxaf:
nice catch

PR #517, opened 2025-04-04
Use shell parsing to split fzf-command into words

andelink:
Small change only to apply the `(z)` flag when expanding the `fzf_command` parameter. This allows for proper command lists to be provided e.g.

```shell
zstyle ':fzf-tab:*' fzf-command 'env SOME_VAR=foo fzf'`
```

Without this change, the above fails with:

```shell
zsh: command not found: env SOME_VAR =foo fzf ...
```


andelink:
Hi @Aloxaf, wondering if you can take a quick look at this. Currently the `fzf-command` setting is quite limited and this PR is a super small change to fix that. Let me know if there is anything I can do differently here. Thanks!

PR #531, opened 2025-07-02
fix: solving msys2 regressions

### LINKED ISSUE #530: [BUG] msys2 path problem reappears
#### Describe the bug
See #241
I'm sure it's a regression bug, because you fixed it before with [this code](https://github.com/Aloxaf/fzf-tab/blob/2abe1f2f1cbcb3d3c6b879d849d683de5688111f/fzf-tab.zsh#L114C1-L114C12), but msys2 released the [destructive update](https://github.com/msys2/msys2.github.io/blob/983b1bdaee36524de4cf0d075b20488e7f6b4d2a/web/news.md?plain=1#L78) that invalidated this fix

I can make sure:
- [ ] this is the minimal zshrc which can reproduce this bug
- [ ] fzf-tab is loaded after plugins which will wrap <kbd>Tab</kbd>, like [junegunn/fzf/completion.zsh](https://github.com/junegunn/fzf/blob/master/shell/completion.zsh)
- [ ] fzf-tab is loaded before zsh-autosuggestions, zsh-syntax-highlighting and fast-syntax-highlighting.

#### To Reproduce
1. Press <kbd>Tab</kbd>
2. See error‘unsupported key: C:/msys64/’

#### Expected behavior
fzf Normal occurrence.

#### Environment:
 - OS: [windows 11]
 - zsh version: [5.9]

#### Minimal zshrc
If applicable, add a minimal zshrc to help us analyze.

#### Log
If applicable, use `C-x .` to trigger completion and provide the log.

If there are only three lines in your log, please make sure your fzf-tab is loaded with the correct order (see the checklist above).

hour-keeper:
fix #530

### DIFF
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -111,7 +111,7 @@ builtin unalias -m '[^+]*'
   -ftb-generate-complist # sets `_ftb_complist`
 
   -ftb-zstyle -s continuous-trigger continuous_trigger || {
-    [[ $OSTYPE == msys ]] && continuous_trigger=// || continuous_trigger=/
+    [[ $OSTYPE == cygwin ]] && continuous_trigger=// || continuous_trigger=/
   }
 
   case $#_ftb_complist in


filips123:
Unfortunately, this PR now breaks Git For Windows shells, where `OSTYPE` is still set to `msys`.

hour-keeper:
> Unfortunately, this PR now breaks Git For Windows shells, where `OSTYPE` is still set to `msys`.

Maybe you're talking about [the msys2 fork of gitforwindows](https://gitforwindows.org/install-inside-msys2-proper)? I have the same variables as msys2 for this one without the git-extra package installed.

filips123:
I'm using Git for Windows version of MSYS2, the one with comes with normal installation of Git for Windows. And although it has Bash installed by default, I managed to add Zsh there.

hour-keeper:
> I'm using Git for Windows version of MSYS2, the one with comes with normal installation of Git for Windows. And although it has Bash installed by default, I managed to add Zsh there.

I recommend you try [this](https://gitforwindows.org/install-inside-msys2-proper). gitforwindows packages are sometimes relatively old.

hour-keeper:
> I'm using Git for Windows version of MSYS2, the one with comes with normal installation of Git for Windows. And although it has Bash installed by default, I managed to add Zsh there.

https://github.com/msys2/MINGW-packages/pull/26470

PR #553, opened 2026-01-03
New style option to fzf-tab group headers: `active-group-style`

alberti42:
Hi Aloxaf!

This PR adds a new style option to fzf-tab group headers: `active-group-style`.

Let me know if you'd prefer the style tokens to support additional SGR options in the future (e.g., reverse, italics).

If you like the PR and plan to make it upstream, I am glad to help you update the documentation. I did not change it so far, as I prefer to receive your feedback first.

Best regards,
Andrea

### New: `active-group-style` zstyle

You can now control how the *active* group label is highlighted in the header:

```
# underline and make bold the active group label
zstyle ':fzf-tab:*' active-group-style underline,bold
```

Supported values:

* `bold`
* `underline`
* `underline,bold` (or `bold,underline`)

The style is normalized internally, so users can write `underline + bold`, `BOLD,Underline`, etc.

### Bug fix: active group label is highlighted on the first render

Previously, the active group label would not be highlighted on the initial fzf-tab menu render, and only became bold after the first group switch. With this change, the active group label is styled correctly from the very first render (defaulting to the first group, or a persisted selection when available).

### Implementation notes

* Reloads and the initial render share the same code path (`ftb-switch-group`), improving consistency.

### Small improvement: GNU grep / `-P` detection

Separate commit 8dddc5944a20df4cd21e136b63a84969dae38b2d: As a minor robustness/portability improvement, the code now resolves an absolute grep path via `$commands[grep]` / `$commands[ggrep]` and only uses the fast `grep -P` path when GNU grep is available (`--version` succeeds). This avoids issues with aliases and BSD grep on macOS, while falling back to a pure-zsh extraction when needed.

### Backwards compatibility

* Default behavior remains `bold` if `active-group-style` is not set.
* Users who don’t care about the feature should see no behavior change other than the initial highlight bug fix.

### AI Assistance Disclaimer

This pull request, including the implementation of the patch and the descriptive text, was developed with the assistance of an AI model (ChatGPT 5.2). The process was guided by a human developer who provided the specific logic, identified the target locations for surgical code changes, and directed the content and structure of this note. The final code and functionality have been thoroughly tested and validated by humans on macOS (Apple Silicon) and Linux (Ubuntu 22.04) to ensure correctness and stability.

### DIFF
--- a/lib/-ftb-fzf
+++ b/lib/-ftb-fzf
@@ -24,7 +24,7 @@ ftb_preview_init=${ftb_preview_init//__FTB_WORDS_DUMP__/$(typeset -p words)}
 ftb_preview_init+=$'\n'
 
 local default_binds=tab:down,btab:up,change:top,ctrl-space:toggle,bspace:backward-delete-char/eof,ctrl-h:backward-delete-char/eof
-local fzf_command fzf_flags fzf_preview debug_command tmp switch_group active_group_style fzf_pad fzf_min_height binds 
+local fzf_command fzf_flags fzf_preview debug_command tmp switch_group active_group_style fzf_pad fzf_min_height binds
 local ret=0
 
 -ftb-zstyle -s fzf-command fzf_command || fzf_command=fzf
--- a/lib/ftb-switch-group
+++ b/lib/ftb-switch-group
@@ -53,7 +53,7 @@ if (( $#group_sgr_prefixes > 0 )) && [[ -f $tmp_dir/current-group.$pid ]]; then
   (( current > $#group_sgr_prefixes )) && current=1
   (( current <= 0 )) && current=$#group_sgr_prefixes
 fi
-echo $current > $tmp_dir/current-group.$pid
+print -r -- $current > $tmp_dir/current-group.$pid
 
 # configure style of active header
 local sgr_on='' sgr_off=''


alberti42:
Hi Aloxaf!

I added an additional commit, 05bf9576131ea7b78a00a747b0afa364f6d52aad. This could have been a separate PR.

It is about writing a more robust and easier-to-maintain `ftb_preview_init` function that does not require hacks (like dynamical string replacement) and escaping all its fragile characters. The new code is fully compatible with the previous one.

You can now open `ftb_preview_init` as a separate file with your favorite editor supporting `zsh` syntax highlighting. This helps substantially to write better code and avoid bugs that can sneak in, possibly due to incorrectly matched parentheses.

BTW: Thanks for your continuous support to this wonderful utility!

Andrea

intelfx:
Nice, I like it. I pulled this into my own fork and implemented some more styles.

> Previously, the active group label would not be highlighted on the initial fzf-tab menu render, and only became bold after the first group switch. With this change, the active group label is styled correctly from the very first render (defaulting to the first group, or a persisted selection when available).

How can you see this in action? I don't see any highlighting on the first render, but I assume that's because normally Zsh's completion menu defaults to "all groups" — do you use some additional configuration such that the first/previously used group is selected from the start?

alberti42:
I also was surprised that it was not showing it on the first start. But I remember I found the way to refactor the code to make it work on the first start. Doesn't it work for you as well?

This is how it looks for me on first start.

Bildschirmfoto 2026-02-04 um 18 15 48

You could try to run https://github.com/alberti42/fzf-tab-fork/tree/integrated. This is my current version. It is essentially upstream/master + the styling of groups.

I have no clue why GitHub says that this branch is 29 commits ahead of the upstream/master. I don't think this true... probably it counts in a different way as I count them.

Then you can check if it works for you. Hopefully we will polish this and with @Aloxaf's feedback, we will reach a good satisfactory level that can be made upstream one day.

Aloxaf:
Thanks for your contribution! I will review it in this weekend.

lib/-ftb-fzf
```diff
@@ -87,7 +146,9 @@ if [[ "$use_fzf_default_opts" == "yes" ]]; then
   fzf_default_opts=$FZF_DEFAULT_OPTS
 fi
 
-FZF_DEFAULT_OPTS=$fzf_default_opts SHELL=$ZSH_NAME ${(z)fzf_command} \
+local -a initial_command=($commands[zsh] -f $FZF_TAB_HOME/lib/ftb-switch-group $$ $header_lines $normalized_active_group_style $tmp_dir 0)
+
+FZF_DEFAULT_OPTS=$fzf_default_opts SHELL=$ZSH_NAME $fzf_command \
```
Aloxaf: should be `${(z)fzf_command}`

lib/ftb-switch-group
```diff
@@ -4,35 +4,88 @@ emulate -L zsh -o extended_glob
...
+  fi
+}
+
+# --- group marker detection (fzf-tab groups) ---------------------------------
+# When groups exist, fzf-tab prefixes each entry with a group SGR like $'\e[94m'.
+# When no groups exist (often file lists), entries may start with $'\e[0m'
+# (reset for LS_COLORS). Treat that as "no groups".
+
+local -Ua group_sgr_prefixes
+if [[ -n $grep_cmd ]]; then
```
Aloxaf: Use $#list > 10000 as condition to avoid unnecessary external command call

lib/-ftb-fzf
```diff
@@ -87,7 +146,9 @@ if [[ "$use_fzf_default_opts" == "yes" ]]; then
   fzf_default_opts=$FZF_DEFAULT_OPTS
 fi
 
-FZF_DEFAULT_OPTS=$fzf_default_opts SHELL=$ZSH_NAME ${(z)fzf_command} \
+local -a initial_command=($commands[zsh] -f $FZF_TAB_HOME/lib/ftb-switch-group $$ $header_lines $normalized_active_group_style $tmp_dir 0)
```
Aloxaf:
```diff
- local -a initial_command=($commands[zsh] -f $FZF_TAB_HOME/lib/ftb-switch-group $$ $header_lines $normalized_active_group_style $tmp_dir 0)
+ local -a initial_command=($(z)reload_command 0)
```

alberti42:
@Aloxaf 

Thanks for reviewing the PR. I think I could cover all the points you have raised. Some of them have much simplified the code! Let me know if you still need anything from my side.

Aloxaf:
LGTM

alberti42:
Awesome. Happy that the new feature is upstream!

Let me know if you need me to write a draft for the documentation of the new features.

QuarticCat:
> Previously, the active group label would not be highlighted on the initial fzf-tab menu render, and only became bold after the first group switch. With this change, the active group label is styled correctly from the very first render (defaulting to the first group, or a persisted selection when available).

Previously, the first menu showed all candidates, not just candidates from the first group. After this PR, only candidates from the first group will show. I believe the previous behavior is correct.

alberti42:
Yes, we never discussed this with @Aloxaf. It is a change of behavior by design. I should have brought it up, but I had forgotten it.

The previous behavior was confusing to me because it showed all options, however, without clearly stating it.

I would suggest the following:

-1 We add a configuration option for fzf-tab (zstyle...) that allows the user to choose whether he/she likes to have a first group showing all options together. Default: true (to be consistent with the previous behavior).
-2 If the configuration for all options is true, we add a new group in the beginning called "All"

Do you think it would be a viable solution?

PR #554, opened 2026-01-04
Feat: support _approximate to correctly show correction options

### LINKED ISSUE #56: Support _approximate completer
`_approximate` will "hook" compadd as well.

A way to support it is to unfunction compadd when call _approximate and re-define it when call other completer/functions (_complete, _description).

### LINKED ISSUE #470: [BUG] FZF tab completion does not activate on similar file names
#### Describe the bug
When trying to complete entries that have the same starting substring, fzf completion will not be triggered, and thus only a regular zsh menu completion is possible. This is unlike [lincheney's solution](https://github.com/lincheney/fzf-tab-completion) which works flawlessly, but lacks features that I need.

I can make sure:

#### To Reproduce
1. Create an empty directory, and create in it two files with identical starting names. For instance:
```sh
mkdir test
touch aaae
touch aaad
```
2. Go inside the directory, type `cat` and press <kbd>Tab</kbd>

#### Expected behavior
FZF is opened and I am able to complete what I want.

#### Screenshots
image

How it should be:
image

#### Environment:
 - OS: Arch Linux
 - zsh version: 5.9

#### Minimal zshrc
```zsh
# Completion engine settings
zstyle ':completion:*' completer _oldlist _expand _complete _ignored _match _correct _approximate _prefix
zstyle ':completion:*' matcher-list 'm:{[:lower:][:upper:]}={[:upper:][:lower:]}' 'r:|[._-]=** r:|=**' 'l:|=* r:|=*'
zstyle ':completion:*' menu select

# Load completions engine
autoload -Uz compinit
compinit -d ~/.cache/zcompdump
zinit cdreplay -q

# Load FZF-tab
zinit light Aloxaf/fzf-tab
```

#### Log
`Aloxaf/fzf-tab`
image

`lincheney/fzf-tab-completion`
image

`fzf builtin completion`
image

alberti42:
This PR resolves long-standing issues where `_approximate` (fuzzy matching) and `_correct` would bypass `fzf-tab` and fall back to the standard Zsh menu completion.

Fixes #56
Fixes #470

### The Problem
Support for `_approximate` has been broken due to three specific conflicts between `fzf-tab` and Zsh's internal implementation of the completer:

1.  **Hook Conflict:** `_approximate` checks `if (( ! $+functions[compadd] ))`. Since `fzf-tab` defines `compadd` globally, `_approximate` detects this and skips defining its local wrapper. This prevents the fuzzy globbing logic (e.g., `(#a1)`) from ever initializing.
2.  **Capture Bypass:** Even if the wrapper runs, `_approximate` explicitly calls `builtin compadd`, bypassing the `fzf-tab` capture function (`-ftb-compadd`).
3.  **Early Exit on Unambiguous Prefix:** When `_approximate` generates corrections (e.g., `Docn` -> `Documents`, `Downloads`), Zsh calculates a common unambiguous prefix (`Do`). `fzf-tab` detects this unambiguous prefix and auto-accepts it, exiting before displaying the selection list, even though `_approximate` sets `compstate[list]="list force"`.

### The Solution

This PR implements a robust fix involving dynamic patching and a logic update in the completion handler:

1.  **Dynamic Source Patching:**
    In `enable-fzf-tab`, we dynamically load `_approximate` and create a patched version where `builtin compadd` is replaced with `-ftb-compadd`. This ensures we capture the matches without maintaining a hard-coded copy of the `_approximate` source code.

2.  **Temporary Hook Unregistration:**
    We wrap the execution of `_approximate`. inside this wrapper, we run `unfunction compadd` before calling the patched logic. This satisfies `_approximate`'s internal check, forcing it to define its local fuzzy-matching wrapper (which now calls our capture function). The global hook is restored immediately in an `always` block.

3.  **Respect `compstate[list] force`:**
    Updated `-ftb-complete` to check if `compstate[list]` contains `"force"`. If it does, `fzf-tab` will now display the list even if there is an unambiguous prefix, allowing the user to select between the original input and corrections.

### Verification configuration

Tested with the following configuration:

```zsh
zstyle ':completion:*' completer _complete _match _approximate
zstyle ':completion:*:match:*' original only
zstyle ':completion:*:approximate:*' max-errors 1 numeric
```

**Scenario:**
Type `cd Docn` and press TAB.
**Before:** Zsh standard menu appears below the prompt.
**After:** `fzf-tab` opens, showing `Documents` and `Downloads` (and the original `Docn`) correctly.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019-2024 Aloxaf
+Copyright (c) 2019-2026 Aloxaf
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/fzf-tab.zsh
+++ b/fzf-tab.zsh
@@ -470,8 +470,6 @@ zmodload -F zsh/stat b:zstat
 0="${${(M)0:#/*}:-$PWD/$0}"
 FZF_TAB_HOME="${0:A:h}"
 
-source "$FZF_TAB_HOME"/lib/zsh-ls-colors/ls-colors.zsh fzf-tab-lscolors
-
 typeset -ga _ftb_group_colors=(
   $'\x1b[94m' $'\x1b[32m' $'\x1b[33m' $'\x1b[35m' $'\x1b[31m' $'\x1b[38;5;27m' $'\x1b[36m'
   $'\x1b[38;5;100m' $'\x1b[38;5;98m' $'\x1b[91m' $'\x1b[38;5;80m' $'\x1b[92m'
@@ -481,6 +479,13 @@ typeset -ga _ftb_group_colors=(
 # init
 () {
   emulate -L zsh -o extended_glob
+
+  # Ensure we don't keep a stale value around.
+  # If the module is loaded, the variable is valid; unsetting would lose it
+  # since zmodload won't reload an already-loaded module.
+  if ! zmodload -e aloxaf/fzftab; then
+    unset FZF_TAB_MODULE_VERSION 2>/dev/null
+  fi
 
   if (( ! $fpath[(I)$FZF_TAB_HOME/lib] )); then
     fpath+=($FZF_TAB_HOME/lib)
@@ -497,7 +502,7 @@ typeset -ga _ftb_group_colors=(
 
   if [[ -n $FZF_TAB_HOME/modules/Src/aloxaf/fzftab.(so|bundle)(#qN) ]]; then
     module_path+=("$FZF_TAB_HOME/modules/Src")
-    zmodload aloxaf/fzftab
+    zmodload aloxaf/fzftab # if this fails, we fall back to ls-colors.zsh below
 
     if [[ $FZF_TAB_MODULE_VERSION != "0.2.2" ]]; then
       zmodload -u aloxaf/fzftab
@@ -510,6 +515,9 @@ typeset -ga _ftb_group_colors=(
       fi
     fi
   fi
+
+  # Only needed when the module is not available (or failed to load).
+  [[ $FZF_TAB_MODULE_VERSION = "0.2.2" ]] || source "$FZF_TAB_HOME"/lib/zsh-ls-colors/ls-colors.zsh fzf-tab-lscolors
 }
 
 enable-fzf-tab
--- a/lib/-ftb-fzf
+++ b/lib/-ftb-fzf
@@ -70,7 +70,7 @@ if (( ! use_tmux_popup )); then
   echoti cnorm >/dev/tty 2>/dev/null
 fi
 
-cat > $tmp_dir/completions.$$
+command cat > $tmp_dir/completions.$$
 
 local dd='gdd'
 if (( ${+commands[$dd]} == 0 )) ; then
--- a/lib/-ftb-generate-complist
+++ b/lib/-ftb-generate-complist
@@ -86,6 +86,7 @@ fi
 local LC_ALL=C
 zstyle -T ":completion:$_ftb_curcontext" sort
 if (( $? != 1 )); then
+  local LC_ALL=C
   if (( colorful )); then
     # if enable list_colors, we should skip the first field
     if [[ ${commands[sort]:A:t} != (|busybox*) ]]; then


alberti42:
I have pushed a refinement to the `PREFIX` sanitization logic (commit 8f6531f40f1189e6227429367b129ef3d45a8221).

Instead of globally stripping the `(#a<num>)` pattern, I switched to **anchored substitution** (`/#`).

**Reasoning:**

Upon reviewing the `_approximate` source code and Zsh documentation ([Section 14.8.5](https://zsh.sourceforge.io/Doc/Release/Expansion.html)), `_approximate` only injects the approximation flag in two specific locations:
1.  **At the very start:** `PREFIX="(#a${_comp_correct})$PREFIX"`
2.  **After a leading tilde:** `PREFIX="~(#a${_comp_correct})${PREFIX[2,-1]}"`

The previous global replacement was too aggressive. If a user had a file literally named `(#a1)test` and typed it explicitly, the global replace would strip the literal characters from the query.

The new logic strictly targets only the flags injected by Zsh, preserving the integrity of the user's actual input in all other cases:

```zsh
# 1. Strip flag at start
expanded=${expanded/#\(\#a[0-9]##\)/}
# 2. Strip flag after leading tilde (e.g. ~(#a1)/foo -> ~/foo)
expanded=${expanded/#\~\(\#a[0-9]##\)/~}
```

alberti42:
### Follow-up: apply fzf-tab selections on non-zero return, and keep list output stable

The commit 2b5e3974803a14ee8a1b427a2574157b94db780d is a follow-up to this PR #554 (_approximate integration). After that change, some completion flows run through the `approximate/correct` completer, which can legitimately return a non‑zero status when it does not insert a match (ambiguous or list‑only). `fzf-tab` still captures valid choices in those cases, but the old gate (`ret == 0 && _ftb_finish == 0`) prevented `_fzf-tab-apply` from running.

That’s why completions like `cd $ZINIT[PLUGI<Tab>` showed choices but would not insert `PLUGINS_DIR` (notably in the “association key” group).

#### Fix

Apply selections based on `_ftb_choices` (and `_ftb_finish`) instead of the original widget’s return status. This treats a non‑zero `ret` as a normal completion outcome rather than a failure.

Additionally, when completion produces no matches but emits a warning message (via `compadd -x`), the widget can still return success while leaving `compstate[list]` set (e.g., `ambiguous rows`). ZLE will then try to redraw the list after the widget finishes, which reprints the warning. We now clear `compstate[list]`/`compstate[insert]` in that success/no-choice path to prevent the extra redraw. This fixes the `warnings` test case that does:

```zsh
comptesteval 'zstyle ":completion:*:warnings" format "<WARN>%d</WARN>"'
comptest $': asd\t'
```

and expects a single `<WARN>` line.

This PR also includes the `LC_ALL=C` sorting guard that was already proposed separately in Aloxaf/fzf-tab#557. It is included here because the completion ordering in this change depends on C collation to keep list output stable across locales and to keep the test expectations consistent.

#### Quick validation

After preparing a few dummy test variables:

```zsh
typeset -A test_variable
test_variable[my_first_test]='Hello'
test_variable[my_second_test]='World'
```

then `echo $test_variable[my<Tab>` → selecting `my_first_test` now inserts correctly.

No behavioral change for normal completion paths; only the apply gate and list output are corrected, and sort order is stabilized.

alberti42:
Hi @Aloxaf,

I folded the small but important fix from #558 into this PR (commit 3e966d47b1ce8c4e8c55fc79730ad35e645d66b9). The changes touch the same area, so it felt simpler to review together.

I believe the PR is ready now and all tests pass on my side. If you can take a look and try it locally, I’d appreciate any feedback.

Thanks a lot!
Andrea

alberti42:
Hi @Aloxaf,

I wanted to ask you if you have had the time to take a look at the PR? I put many hours into it, and I think this solves a long-standing problem, which can be very valuable for many people.

I have been testing it on my setup for the last weeks, and everything works well.

I can understand you are very busy and probably have no time to carefully review the PR. But would you mind merging the changes locally onto your computer... so that you can already get a first impression? I would love to hear some feedback like, "Yup, it seems to work," or "There are still things to fix"... This would help a lot, even if it is not a merge upstream, which may take longer before you are confident about the new change.

Big love ❤️ for this little tool, which I use every day.

Best
Andrea

Aloxaf:
Sorry for the late reply.

Thanks for your work, this is awesome!

alberti42:
Awesome, I am glad it worked. I am now running with the upstream version. There was only one last PR that was about groups.. but it less critical and we can discuss it separately to understand what you expect/wish to have. For me, "groups" were always very hard to visually distinguish with the current styling and the color switching was confusing my eyes...

but I am very happy the other PR landed.

By the way, I programmed over the weekend a crazy little zsh plugin https://github.com/alberti42/Zsh-Opencode-Tab Take a look. Maybe it is interesting for you too. If you have feedback, drop some lines in my plugin issue section.

PR #583, opened 2026-07-03
fix: conditionally pass -Q to compadd based on original completion call

### LINKED ISSUE #503: [Q] [carapace + fzf-tab] adds an extra backslash to file completion.
When using carapace in conjunction with fzf-tab, it inadvertently adds an extra backslash to escape spaces or parentheses.

```
: mv Hello\\ One\\ \\(Two\\)\\ \\[Three\\].mp3 test.mp3
zsh: no matches found: \(Two\)\
```

However, when using `**` and pressing tab, the command is executed successfully.
```
mv Hello\ One\ \(Two\)\ \[Three\].mp3
````

Is there a way to resolve this issue either by updating the fzf configuration or the carapace configuration?

As per the developer of Carapace
> Carapace does custom quoting with compadd -Q.
> Fzf-tab probably does that as well (or adds them without -Q) without acknowledging they are already quoted.

### PR DESCRIPTION
**Pull request by nirvana6**
## Problem

When fzf-tab re-inserts a selected match via `compadd`, it unconditionally uses the `-Q` flag ("do not quote metacharacters") via the fallback `"${args[@]:--Q}"`. This breaks path completions containing spaces when the original completer (e.g., `_files`) did not use `-Q` — the literal space is inserted unescaped and treated as a word boundary by zsh.

This is also a known source of conflict with **carapace** (issue #503), which passes pre-quoted words to compadd with `-Q`. fzf-tab should preserve the original completer's decision rather than overriding it.

## Changes

Two call sites in `fzf-tab.zsh` where `compadd` is called with the unconditional `-Q` fallback:

1. **`_fzf-tab-apply`** (line 259): normal match insertion after fzf selection
2. **`-ftb-complete`** (line 209): the print-query / raw-input insertion path

Both changed from:
```zsh
builtin compadd "${args[@]:--Q}" -Q -- "$word"
```
to:
```zsh
if (( ${args[(I)-Q]} )); then
  builtin compadd -Q -- "$word"
else
  builtin compadd -- "$word"
fi
```

## Rationale

- When the original completer uses `-Q`, the word is pre-quoted → preserve `-Q`
- When the original completer does NOT use `-Q`, the word is raw → let compadd handle quoting normally
- This matches zsh's completion contract: pre-quoted words come with `-Q`, raw words come without

## Testing

- [x] Native `_files` path completion with spaces (e.g., Chinese filenames with spaces) now works correctly
- [x] No regression on completions without special characters
- [x] Compatible with carapace completions (carapace passes `-Q`, so behavior is unchanged)

Fixes #503


nirvana6:
Closing — the fix does not fully resolve the issue.

Issue #509, opened 2025-02-11
[BUG] use-fzf-default-opts has no effect in ftb-tmux-popup


psicho2000:
#### Describe the bug

* use-fzf-default-opts works flawlessly outside of tmux
* within tmux, use-fzf-default-opts seems to be ignored

I can make sure:
- [x] I am using the latest version of fzf-tab (using commit #6aced3f from master)
- [x] this is the minimal zshrc which can reproduce this bug
- [x] fzf-tab is loaded after `compinit`
- [x] fzf-tab is loaded after plugins which will wrap <kbd>Tab</kbd>, like [junegunn/fzf/completion.zsh](https://github.com/junegunn/fzf/blob/master/shell/completion.zsh)
- [x] fzf-tab is loaded before zsh-autosuggestions, zsh-syntax-highlighting and fast-syntax-highlighting.

#### To Reproduce
1. Enter tmux
2. Type 'cd'<kbd>Tab</kbd>

#### Expected behavior
Inside tmux (using `fzf-command ftb-tmux-popup`), `use-fzf-default-opts` should be respected.

#### Screenshots
Reproduction outside of tmux works as expected (see custom pointer `x`):
![Image](https://github.com/user-attachments/assets/fbaab679-b61d-4f66-bfe7-336d2fd6789c)

Within tmux, `FZF_DEFAULT_OPTS` are not respected (custom pointer is not applied):
![Image](https://github.com/user-attachments/assets/366161cb-c6de-4220-8de6-3f9e00cfa622)

#### Environments
Environment A
 - OS: macOS. Sequoia 15.3.1
 - zsh version: zsh 5.9 (arm-apple-darwin24.2.0)

Environment B
 - OS: Ubuntu 24.04.1 LTS
 - zsh version: zsh 5.9 (x86_64-ubuntu-linux-gnu)

#### Minimal zshrc
```
# path to fzf
PATH="$PATH:/opt/homebrew/bin"

export FZF_DEFAULT_OPTS='--pointer x'
zstyle ':fzf-tab:*' fzf-command ftb-tmux-popup
zstyle ':fzf-tab:*' use-fzf-default-opts yes

autoload -Uz compinit && compinit -u -C
# path to fzf-tab from git clone
source ~/bin/fzf-tab/fzf-tab.zsh
```

#### Log
As completion itself works, completion log won't give any insights here.

Issue #455, opened 2024-07-13
[BUG] Using `--tmux` in an exported `FZF_DEFAULT_OPTS` breaks `ftb-tmux-popup`


andrewbraxton:
#### Describe the bug

Using `--tmux` in an exported `FZF_DEFAULT_OPTS` breaks `ftb-tmux-popup`. Weird thing is that it seems to stay broken even if you `unset FZF_DEFAULT_OPTS`.

I can make sure:
- [x] I am using the latest version of fzf-tab
- [x] this is the minimal zshrc which can reproduce this bug
- [x] fzf-tab is loaded after `compinit`
- [x] fzf-tab is loaded after plugins which will wrap <kbd>Tab</kbd>, like [junegunn/fzf/completion.zsh](https://github.com/junegunn/fzf/blob/master/shell/completion.zsh)
- [x] fzf-tab is loaded before zsh-autosuggestions, zsh-syntax-highlighting and fast-syntax-highlighting.

#### To Reproduce
1. Enter tmux
1. Type `ls` (or anything)
2. Press <kbd>Tab</kbd>
3. Popup flashes briefly and disappears

#### Expected behavior
Popup appears normally and allows selection

#### Environment:
 - OS: macOS Sonoma 14.5
 - zsh version: 5.9

#### Minimal zshrc

```
autoload -Uz compinit
compinit

export FZF_DEFAULT_OPTS='--tmux'

source ~/.fzf-tab/fzf-tab.plugin.zsh
zstyle ':fzf-tab:*' fzf-command ftb-tmux-popup
```

cameronbrill:
I am also running into this issue

cameronbrill:
interestingly, if I remove `export FZF_DEFAULT_OPTS='--tmux'` from my `zshrc`, start a tmux session, add `export FZF_DEFAULT_OPTS='--tmux'` to my zshrc, then run `zsh` to restart the shell, both ftb-tmux-popup and the fzf tmux popup work as intended

andrewbraxton:
Yeah, I ran a bunch of the same experiments and got similar confusing results.

itahol:
I have the same problem, but I couldn't resolve it even when trying your trick. Did you figure this out?

itahol:
My current workaround is editing `ftb-tmux-popup` so it runs fzf without "--tmux" in its opts, like so:
```zsh
[...]

PR #588, opened 2026-08-18
fix(ftb-tmux-popup): strip --tmux/--popup from FZF_DEFAULT_OPTS

### LINKED ISSUE #587: [BUG] ftb-tmux-popup nests when FZF_DEFAULT_OPTS contains --tmux/--popup → DA ?1;2;4c leak
#### Describe the bug
`ftb-tmux-popup` (`lib/ftb-tmux-popup:27`) only strips `--height*`/`--layout*` from `fzf_opts`. With `zstyle ':fzf-tab:*' use-fzf-default-opts yes` and `export FZF_DEFAULT_OPTS='--tmux'` (or `--popup`), the popup does a nested popup: outer `tmux popup -E` (from `ftb-tmux-popup`) + inner `fzf --tmux` popup. Result: popup flashes and disappears, inserting the query text instead of showing matches.

History: #455 reported `--tmux` in `FZF_DEFAULT_OPTS` breaks `ftb-tmux-popup`. #509 fixed it by propagating `FZF_DEFAULT_OPTS` into the popup (`01dad759` changed `FZF_DEFAULT_OPTS=''` → `FZF_DEFAULT_OPTS=${(qq)FZF_DEFAULT_OPTS}`), which re-exposes the nested-popup case.

I can make sure:

#### To Reproduce
1. `export FZF_DEFAULT_OPTS='--tmux'`
2. `zstyle ':fzf-tab:*' fzf-command ftb-tmux-popup` + `zstyle ':fzf-tab:*' use-fzf-default-opts yes`
3. Type `ls ` and press <kbd>Tab</kbd> inside tmux (with an attached client)

#### Expected behavior
fzf-tab shows the file matches inside the single `tmux popup`; query is the typed prefix, no nested popup.

#### Screenshots
No screenshot captured.

#### Environment:
 - OS: Arch Linux
 - zsh version: 5.9
 - tmux: 3.7b
 - fzf: 0.74.2
 - fzf-tab: `fc6f0dc` (master)

#### Minimal zshrc
```zsh
autoload -Uz compinit; compinit
export FZF_DEFAULT_OPTS='--tmux'
source ~/.local/share/zinit/plugins/Aloxaf---fzf-tab/fzf-tab.plugin.zsh
zstyle ':fzf-tab:*' fzf-command ftb-tmux-popup
zstyle ':fzf-tab:*' use-fzf-default-opts yes
```

#### Log
Verified with the minimal zshrc above in tmux 3.7b with an attached client (`script -qec 'tmux attach'`):
- `ls <TAB>` → popup flashes and disappears; buffer becomes `ls ls` (the query text inserted, no match list shown).
- The generated popup script contains `env FZF_DEFAULT_OPTS='--tmux' ... fzf ...` — fzf reads `--tmux` from the env and attempts a second (nested) popup inside the outer `tmux popup -E`, which fails and closes.

Fix verified locally (minimal zshrc + tmux attached client; popup stays open, `env FZF_DEFAULT_OPTS=''` in generated script):
```diff
 # lib/ftb-tmux-popup: strip --tmux/--popup from FZF_DEFAULT_OPTS before it
 # reaches fzf inside the popup (bare and =value forms; keep --height/--layout)
+local _ftb_fdo=${FZF_DEFAULT_OPTS:-}
+if [[ -n $_ftb_fdo ]]; then
+  local -a _ftb_fdo_opts=(${(z)_ftb_fdo}) _ftb_fdo_kept=() _ftb_fdo_opt
+  for _ftb_fdo_opt in $_ftb_fdo_opts; do
+    case $_ftb_fdo_opt in
+      --tmux|--popup|--tmux=*|--popup=*) ;;
+      *) _ftb_fdo_kept+=($_ftb_fdo_opt) ;;
+    esac
+  done
+  _ftb_fdo=${(j: :)_ftb_fdo_kept}
+fi
-echo -E "env FZF_DEFAULT_OPTS=${(qq)FZF_DEFAULT_OPTS} ..." > $tmp_dir/fzf-$$
+echo -E "env FZF_DEFAULT_OPTS=${(qq)_ftb_fdo} ..." > $tmp_dir/fzf-$$
```
Note: the earlier revision proposed stripping `--tmux`/`--popup` from `fzf_opts` (`fzf_opts=(${${${${fzf_opts/--height*}/--layout*}/--tmux*}/--popup*})`). That is NOT sufficient — fzf-tab's own `fzf_opts` never contains `--tmux`/`--popup`; the nesting comes from the `FZF_DEFAULT_OPTS` env propagated into the popup script (upstream #509). The env must be filtered, not `fzf_opts`. (The 4× nested `${}` can also be written as one alternation: `fzf_opts=(${fzf_opts/--(height|layout|tmux|popup)*})` — brace form `--{a,b,c}` does not expand inside `${}` substitution.)

HaleTom:
Fixes #587

## Problem

`ftb-tmux-popup` (`lib/ftb-tmux-popup:27`) only strips `--height*`/`--layout*` from `fzf_opts`. With `zstyle ':fzf-tab:*' use-fzf-default-opts yes` and `export FZF_DEFAULT_OPTS='--tmux'` (or `--popup`), the popup does a nested popup: outer `tmux popup -E` (from `ftb-tmux-popup`) + inner `fzf --tmux` popup. Result: popup flashes and disappears, inserting the query text instead of showing matches.

History: #455 reported `--tmux` in `FZF_DEFAULT_OPTS` breaks `ftb-tmux-popup`. #509 fixed it by propagating `FZF_DEFAULT_OPTS` into the popup (`01dad759` changed `FZF_DEFAULT_OPTS=''` → `FZF_DEFAULT_OPTS=${(qq)FZF_DEFAULT_OPTS}`), which re-exposes the nested-popup case.

## Fix

Strip `--tmux`/`--popup` (bare and `=value` forms) from the `FZF_DEFAULT_OPTS` env value passed to fzf inside the popup. Keep `--height`/`--layout` and all other opts.

```zsh
# ${array:#pattern} filters the array as a whole and drops matching elements.
local -a _ftb_fdo_opts=(${(z)FZF_DEFAULT_OPTS})
local _ftb_fdo=${(j: :)_ftb_fdo_opts:#--(tmux|popup)*}
```

Note: filtering `fzf_opts` is NOT sufficient — fzf-tab's own `fzf_opts` never contains `--tmux`/`--popup`; the nesting comes from the `FZF_DEFAULT_OPTS` env propagated into the popup script (upstream #509). The env must be filtered.

## Tests

21 new ztst cases in `test/ftb-tmux-popup.ztst` covering:
- bare `--tmux`, `--popup`
- `--tmux=center,50%`, `--popup=bottom,30%`
- mixed with `--height=40% --color=fg:1` (others kept)
- bare `--tmux` does not eat the next arg
- empty input
- no tmux/popup untouched
- `--popup=` stripped among others
- `--height 40%` separate-arg kept, `--tmux` stripped
- both `--tmux` and `--popup` stripped
- both `=value` forms stripped
- tmux/popup at ends, middle kept
- height/layout kept, `tmux=` stripped
- all four forms stripped
- stripped result has zero elements (no `''` placeholder) — via `$#_r` on `(z)`-split result
- kept args round-trip through `%q` unchanged
- quoted arg with space (`--'has space'`) stays one element after `(z)` split; `%q` round-trips as `--\'has space\'` proving no `''` placeholder is introduced

Verified: `zsh +Z -f ./test/ztst.zsh test/ftb-tmux-popup.ztst` → all tests successful. Existing `test/fzftab.ztst` still passes.

## Verification

Minimal zshrc + tmux 3.7b with attached client:
- Before: `ls <TAB>` → popup flashes and disappears; buffer becomes `ls ls` (query text inserted, no match list).
- After: popup stays open, `env FZF_DEFAULT_OPTS=''` in generated script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 *.zwc
+.zcompdump
--- a/lib/ftb-tmux-popup
+++ b/lib/ftb-tmux-popup
@@ -29,16 +29,9 @@ fzf_opts=(${${fzf_opts/--height*}/--layout*})
 # Strip --tmux/--popup from FZF_DEFAULT_OPTS before it reaches fzf inside the
 # popup: with use-fzf-default-opts yes, fzf would otherwise open a second
 # (nested) popup inside the outer `tmux popup -E` (see #455, #509).
-ftb-tmux-popup-strip-fdo() {
-  emulate -L zsh -o extended_glob
-  local _ftb_fdo=${1:-}
-  if [[ -n $_ftb_fdo ]]; then
-    local -a _ftb_fdo_opts=(${(z)_ftb_fdo})
-    _ftb_fdo=${(j: :)_ftb_fdo_opts:#--(tmux|popup)*}
-  fi
-  print -r -- "$_ftb_fdo"
-}
-local _ftb_fdo=$(ftb-tmux-popup-strip-fdo "$FZF_DEFAULT_OPTS")
+# ${array:#pattern} filters the array as a whole and drops matching elements.
+local -a _ftb_fdo_opts=(${(z)FZF_DEFAULT_OPTS})
+local _ftb_fdo=${(j: :)_ftb_fdo_opts:#--(tmux|popup)*}
 
 # get position of cursor and size of window
 local -a tmp=($(command tmux display-message -p "#{pane_top} #{cursor_y} #{pane_left} #{cursor_x} #{window_height} #{window_width} #{status} #{status-position}"))
--- a/lib/ftb-tmux-popup
+++ b/lib/ftb-tmux-popup
@@ -26,6 +26,20 @@ local ret=0
 local -a fzf_opts=($@)
 fzf_opts=(${${fzf_opts/--height*}/--layout*})
 
+# Strip --tmux/--popup from FZF_DEFAULT_OPTS before it reaches fzf inside the
+# popup: with use-fzf-default-opts yes, fzf would otherwise open a second
+# (nested) popup inside the outer `tmux popup -E` (see #455, #509).
+ftb-tmux-popup-strip-fdo() {
+  emulate -L zsh -o extended_glob
+  local _ftb_fdo=${1:-}
+  if [[ -n $_ftb_fdo ]]; then
+    local -a _ftb_fdo_opts=(${(z)_ftb_fdo})
+    _ftb_fdo=${(j: :)_ftb_fdo_opts:#--(tmux|popup)*}
+  fi
+  print -r -- "$_ftb_fdo"
+}
+local _ftb_fdo=$(ftb-tmux-popup-strip-fdo "$FZF_DEFAULT_OPTS")
+
 # get position of cursor and size of window
 local -a tmp=($(command tmux display-message -p "#{pane_top} #{cursor_y} #{pane_left} #{cursor_x} #{window_height} #{window_width} #{status} #{status-position}"))
 local cursor_y=$((tmp[1] + tmp[2])) cursor_x=$((tmp[3] + tmp[4])) window_height=$tmp[5] window_width=$tmp[6] window_top=0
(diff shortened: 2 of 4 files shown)


HaleTom:
Two incidental test-hygiene changes in commit `dfb27f6` — explaining why they're in this PR:

**1. `test/fzftab.ztst`: `egrep` → `grep -E`**

```diff
-         $(locale -a 2>/dev/null | egrep 'utf8|UTF-8'))
+         $(locale -a 2>/dev/null | grep -E 'utf8|UTF-8'))
```

`egrep` is obsolescent — GNU grep 3.8+ emits `egrep: warning: egrep is obsolescent; using grep -E` on every invocation. The `%prep` block runs `locale -a | egrep` to build the `$langs` array for the `é = ?` locale probe, so every `ztst` run printed the warning to stderr. `grep -E` is the drop-in equivalent (same POSIX ERE semantics), so the locale-detection behaviour is unchanged — it just silences the deprecation noise. This is a test-only change; no runtime code touched.

**2. `.gitignore`: add `.zcompdump`**

```diff
 *.zwc
+.zcompdump
```

The `ztst` harness (`test/ztst.zsh` → `test/comptest`, which sets `export ZDOTDIR=$ZTST_testdir`) generates a `.zcompdump` in the repo root on every test run. It's a generated cache file (60KB, machine-specific), not source. It got accidentally committed twice while iterating on this PR (I had to `git rm --cached` it both times). Adding it to `.gitignore` prevents it from ever being staged again. The existing `*.zwc` entry already ignores the compiled zsh bytecode; `.zcompdump` is the same class of generated artefact.

Both are unrelated to the `--tmux`/`--popup` fix itself but were necessary to keep the test suite clean and warning-free while running the new `test/ftb-tmux-popup.ztst` cases. Happy to split them into a separate commit if preferred.

PR #591, opened 2026-08-24
feat: cancel-group and cancel-group-or-exit bindings

### LINKED ISSUE #590: feat: cancel-group binding
It would be nice to be able to return to the list of all entries once a group is selected.

Even better, I'd like to bind ESC to either cancel the selected group, if one is selected, or exit fzf otherwise.

This was already mentioned by others:

> @Aloxaf Switching and filtering by groups is a nice feature for sure, but I can't find how to cancel the group filter (show all completions) once it's activated.

_Originally posted by @kaznovac in https://github.com/Aloxaf/fzf-tab/issues/183#issuecomment-1492883081_

Thanks for the amazing plugin and persistent work on it ❤️

arnodirlam:
Resolves #590.


lib/-ftb-fzf
```diff
@@ -24,7 +24,8 @@ ftb_preview_init=${ftb_preview_init//__FTB_WORDS_DUMP__/$(typeset -p words)}
 ftb_preview_init+=$'\n'
 
 local default_binds=tab:down,btab:up,change:top,ctrl-space:toggle,bspace:backward-delete-char/eof,ctrl-h:backward-delete-char/eof
-local fzf_command fzf_flags fzf_preview debug_command tmp switch_group active_group_style fzf_pad fzf_min_height binds
+local fzf_command fzf_flags fzf_preview debug_command tmp switch_group cancel_group cancel_group_or_exit
```
arnodirlam, line 27: I was actually thinking to make a `local` line for each variable. That would be easier to diff in the future. WDYT?

lib/ftb-switch-group
```diff
@@ -4,7 +4,19 @@ emulate -L zsh -o extended_glob
 zmodload zsh/mapfile
 
 # receive arguments
-local pid=$1 header_lines=$2 active_group_style=$3 tmp_dir=$4 offset=$5
+local pid=$1 header_lines=$2 active_group_style=$3 tmp_dir=$4 action=$5 value=$6
```
arnodirlam, line 7: Changes the signature of `ftb-switch-group` to `... action=$5 value=$6` so it's more generic.

`value` could instead be named `arg` as well.

